PR #19, opened 2022-03-02
Setup jacoco

Lindt19:
Updated .gitignore and added Jacoco settings according to [https://github.com/sweng-epfl/public/blob/main/sdp/guides/Operations.md](url)
#1


emduc (approved):
Seems all good :)

PR #49, opened 2022-03-09
9-Meds request page and emergency call

emduc:
Implements the basic page for choosing which meds to request when asking for help, along with the possibility to make an emergency call.

### DIFF
--- a/app/src/main/java/com/github/h3lp3rs/h3lp/AwaitHelpActivity.kt
+++ b/app/src/main/java/com/github/h3lp3rs/h3lp/AwaitHelpActivity.kt
@@ -5,7 +5,9 @@ import android.os.Bundle
 import android.widget.TextView
 import android.widget.Toast
 
-
+/**
+ * Activity during which the user waits for help from other user.
+ */
 class AwaitHelpActivity : AppCompatActivity() {
     override fun onCreate(savedInstanceState: Bundle?) {
         super.onCreate(savedInstanceState)
--- a/app/src/main/java/com/github/h3lp3rs/h3lp/AwaitHelpActivity.kt
+++ b/app/src/main/java/com/github/h3lp3rs/h3lp/AwaitHelpActivity.kt
@@ -2,6 +2,8 @@ package com.github.h3lp3rs.h3lp
 
 import androidx.appcompat.app.AppCompatActivity
 import android.os.Bundle
+import android.widget.TextView
+import android.widget.Toast
 
 
 class AwaitHelpActivity : AppCompatActivity() {
@@ -11,6 +13,12 @@ class AwaitHelpActivity : AppCompatActivity() {
 
         val bundle = this.intent.extras
         val array = bundle!!.getStringArrayList(EXTRA_NEEDED_MEDICATION)
+
+        // To be removed once the page is implemented:
+        val displayText = "Selected: $array"
+        val text = findViewById<TextView>(R.id.selected_items_text).apply{
+            text = displayText
+        }
 
         //TODO Launch helper search
         // a: Retrieve nearby users from online database            => loading bar (up to 50%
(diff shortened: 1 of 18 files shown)


app/src/main/res/layout/activity_await_help.xml
```diff
@@ -0,0 +1,36 @@
...
+    xmlns:app="http://schemas.android.com/apk/res-auto"
+    xmlns:tools="http://schemas.android.com/tools"
+    android:layout_width="match_parent"
+    android:layout_height="match_parent"
+    android:id="@+id/await_help_layout"
+    tools:context=".AwaitHelpActivity"
+    android:orientation="horizontal">
+
+    <TextView
+        android:id="@+id/textView3"
```
wiam8, line 12: I would change this id to a more meaningful one to avoid having conflicts with the others.

app/src/main/res/layout/activity_help_parameters.xml
```diff
@@ -0,0 +1,183 @@
...
+        android:textOff="@string/med_pro"
+        android:textOn="@string/med_pro"
+        app:layout_constraintBottom_toTopOf="@+id/help_params_call_button"
+        app:layout_constraintEnd_toEndOf="parent"
+        app:layout_constraintStart_toEndOf="@+id/selectMedsButton4"
+        app:layout_constraintTop_toBottomOf="@+id/selectMedsButton2" />
+
+    <ToggleButton
+        android:id="@+id/selectMedsButton2"
+        android:layout_width="110dp"
```
wiam8, line 88: I suggest adding these dimensions to a separate dimen.xml file so that it is easier to manage and change their respective values later on.

app/src/main/res/layout/activity_await_help.xml
emduc: You're right, this was just a temporary textView until we actually implement this page, but it's always better to be thorough!

app/src/main/res/layout/activity_help_parameters.xml
emduc: Nice catch, this will also come in handy if we change this to dynamic generation later on

app/src/main/java/com/github/h3lp3rs/h3lp/AwaitHelpActivity.kt
```diff
@@ -0,0 +1,32 @@
+package com.github.h3lp3rs.h3lp
+
+import androidx.appcompat.app.AppCompatActivity
+import android.os.Bundle
+import android.widget.TextView
+import android.widget.Toast
+
+
+class AwaitHelpActivity : AppCompatActivity() {
```
AlSchlo, line 9: Needs a proper documentation

app/src/main/java/com/github/h3lp3rs/h3lp/HelpParametersActivity.kt
```diff
@@ -0,0 +1,74 @@
...
+import androidx.appcompat.app.AppCompatActivity
+import android.os.Bundle
+import android.view.View
+import android.view.ViewGroup
+import android.widget.Toast
+import android.widget.ToggleButton
+
+const val EMERGENCY_NUMBER = 112
+
+class HelpParametersActivity : AppCompatActivity() {
```
AlSchlo, line 15: Same here, class documentation can be nice

app/src/main/java/com/github/h3lp3rs/h3lp/MainActivity.kt
```diff
@@ -7,6 +7,7 @@ import android.view.View
 import android.widget.EditText
 
 const val EXTRA_MESSAGE = "KEY"
+const val EXTRA_NEEDED_MEDICATION = "needed_meds_key"
```
AlSchlo, line 10: Keys should be put in a special file later on

AlSchlo (approved):
Nice work, overall classes should have some kind of documentation about what the activity is about. Also, the keys for communication should be centralized in some file. But we will see this later on we get more code.

PR #50, opened 2022-03-09
47 add main page UI

wiam8:
Added the UI of the app's home page. The search bar is not functional yet (it will be done in another sprint). Only the HELP, profile, and button that starts the CPR rate activity are working for now, as we haven't implemented the remaining activities yet. 
The reason the coverage is not at 100% is that I created dummy(empty) activities to be able to test the activity I actually implemented.


app/src/androidTest/java/com/github/h3lp3rs/h3lp/MainPageTestActivity.kt
```diff
@@ -0,0 +1,74 @@
...
+        Intents.intending(IntentMatchers.anyIntent()).respondWith(intentResult)
+
+        onView(withId(R.id.HELP)).perform(click())
+
+        Intents.intended(
+            Matchers.allOf(
+                IntentMatchers.hasComponent(HelpParametersActivity::class.java.name))
+        )
+        Intents.release()
+    }
```
Ahayderi, line 71: There seems to be some amount of code repetition here (which won't scale as we add the other activities and need to test them), you could refactor this testing code into a single method taking as attribute the button id and the class to which the intent should be directed. There would still be the problem that the CPR button will require scrollTo() which the others don't, to tackle this, I have two options:
1. Add a boolean in the general testing function, if it is set, scroll to the button, if not, only click it
2. Create 2 general testing functions, one for methods that will be on the scrollView, one for the ones that don't require scrolling to

Ahayderi (changes requested):
This was great and the only comments I had were somewhat details.
One a more general note: the names of ids and constants in the XML files aren't very consistent (using both under_score and camelCase) but this is not only a problem on your side since it wasn't clear when you actually started your part. To remedy this in the future, I have added conventions for such namings to our shared file but I don't think there is a need to change all of the names in your PR, we can discuss how to check these at our retrospective.

app/src/main/res/layout/activity_main_page.xml
```diff
@@ -0,0 +1,364 @@
...
+                    android:text="@string/skills"
+                    android:textColor="@color/dark_blue"
+                    android:textSize="@dimen/icon_font"
+                    android:textStyle="bold" />
+
+            </LinearLayout>
+        </RelativeLayout>
+
+        <ImageButton
+            android:id="@+id/HELP"
```
emduc, line 353: The id could be a little more precise, given this is the name of the app it might be confusing later on.

eg:


```suggestion
            android:id="@+id/HELP_main_button"
```

app/src/main/res/values/colors.xml
```diff
@@ -7,4 +7,9 @@
     <color name="teal_700">#FF018786</color>
     <color name="black">#FF000000</color>
     <color name="white">#FFFFFFFF</color>
+    <color name="dark_blue">#192841</color>
+    <color name="light_grey">#E7DEDE</color>
+    <color name="app_background">#D6E3EF</color>
+    <color name="blue">#6E85C8</color>
```
emduc, line 13: Storing those in a constant is great. Do you think we should keep the name of the colours or define constants which would indicate what they're used for ?

app/src/main/res/values/dimension.xml
```diff
@@ -0,0 +1,14 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources>
```
emduc, line 2: I defined a file `dimen.xml` to store dimensions, it might a good idea to merge it into yours instead of keeping the 2 versions

emduc (changes requested):
Overall a very nice work, well done ! I commented a few details here and there to finalize it, but it's all good overall. You should also have a look at the Cirrus warnings, some are quite relevant

app/src/main/res/values/colors.xml
wiam8: Once we define the final colors we could add the main ones to the themes file instead!

app/src/main/res/layout/activity_main_page.xml
```diff
@@ -0,0 +1,364 @@
...
+            android:layout_width="150dp"
```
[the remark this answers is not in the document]
wiam8: Will do :)

wiam8:
> This was great and the only comments I had were somewhat details. One a more general note: the names of ids and constants in the XML files aren't very consistent (using both under_score and camelCase) but this is not only a problem on your side since it wasn't clear when you actually started your part. To remedy this in the future, I have added conventions for such namings to our shared file but I don't think there is a need to change all of the names in your PR, we can discuss how to check these at our retrospective.

emduc (approved):
Nice

Ahayderi (approved):
That's great!

PR #52, opened 2022-03-10
11 add cpr rate

Ahayderi:
Implemented the activity giving a user the correct CPR rate through both a sound and a corresponding animation. 

As already noted in the file, I had to comment out animations for the moment since espresso tests freeze if animations are running. I haven't yet found a way to only disable such animations at test time (these aren't the same as the ones that can be disabled in Android developer settings and all the solutions I have found on the Internet only tackle such animations).

Related issues: #42 #40


AlSchlo (approved):
Nice documentation & clean code.
Just 2 files that may be put in the .gitignore.
- .idea/gradle.xml
- graldew (which is an empty file(?))

Lindt19 (approved):
Nothing to add to @AlSchlo 's review. Nice and clean code 👍🏻

PR #80, opened 2022-03-15
77 add side navigation bar

wiam8:
Added a sidebar to the home page with possible features we can implement later. I had to, unfortunately, bring back the action bar to do so.  Depending on whether we prefer to have our own toolbar, it is possible to implement it, I tried to but later realized it is going to be time-consuming, so I can do this on another sprint.
One other issue I ran into after bringing back the action bar is that even though some buttons were still visible on the activities on my machine, they were not on the emulator of Cirrus and so tests were failing. I corrected that by changing the position of the buttons on the page. We should avoid setting the top margin manually to prevent these kinds of issues.


app/src/main/res/values/themes.xml
```diff
@@ -1,8 +1,8 @@
 <resources xmlns:tools="http://schemas.android.com/tools">
     <!-- Base application theme. -->
-    <style name="Theme.H3LP" parent="Theme.MaterialComponents.DayNight.NoActionBar">
+    <style name="Theme.H3LP" parent="Theme.MaterialComponents.DayNight.DarkActionBar">
         <!-- Primary brand color. -->
-        <item name="colorPrimary">@color/purple_500</item>
+        <item name="colorPrimary">@color/light_grey</item>
```
Adrinels, line 5: I think theres no need to change that for now since we have not really decide the grafical chart.

I would anyway propose teal since it is the one in the background image of most of the page

app/src/main/java/com/github/h3lp3rs/h3lp/MainPageActivity.kt
```diff
@@ -2,18 +2,83 @@ package com.github.h3lp3rs.h3lp
...
+        supportActionBar?.setDisplayHomeAsUpEnabled(true)
+
+        navView.setNavigationItemSelectedListener {
+            when (it.itemId) {
+                R.id.nav_profile -> goToProfileActivity(findViewById(R.id.profile))
+                R.id.nav_home -> findViewById<DrawerLayout>(R.id.drawer_layout).closeDrawer(
+                    GravityCompat.START
+                )
+                else -> {
+                    // Toast.makeText(applicationContext, "TODO", LENGTH_SHORT).show()
```
Lindt19, line 54: Maybe remove the toast if not needed

Lindt19 (approved):
Very clean implementation ! Good job 🥇

PR #83, opened 2022-03-16
Adjust codeclimate settings and Fix Jacoco 

emduc:
Adjusts some advanced settings on codeclimate to make it less aggressive and adjust Jacoco settings to account for unit tests. The changes have been made as advised by the TAs in both cases.

### DIFF
--- a/.codeclimate.yml
+++ b/.codeclimate.yml
@@ -1,3 +1,23 @@
+checks:
+  complex-logic:
+    config:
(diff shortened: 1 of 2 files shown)


wiam8 (approved):
Thanks for taking care of this.

Ahayderi (approved):
That's great, thanks for implementing the fix!

PR #84, opened 2022-03-17
Adjust code climate settings

emduc:
Attemps fix on coverage drop


wiam8 (approved):
Hopefully this will work.

Ahayderi (approved):
Great, thanks for the fix

PR #98, opened 2022-03-23
92 make emergency call location dependent

Ahayderi:
This improvement addresses #92 and makes it so that the emergency number called by the app is location-dependent — the number called is the actual ambulance number for the country the user is currently in.

This is implemented by querying the user's location, finding the country that location is in, and then finding the corresponding ambulance phone number using an emergency number database (compiled from different sources to make it as complete as can be).


app/src/androidTest/java/com/github/h3lp3rs/h3lp/LocalEmergencyCallTest.kt
```diff
@@ -0,0 +1,112 @@
+package com.github.h3lp3rs.h3lp
```
wiam8, line 1: Since these tests don't have to run on the emulator, shouldn't they be in the test folder instead of the androidTest folder?

wiam8 (approved):
Good work. By far one of the most concise and detailed documentation I've ever seen!

Would it be possible though to increase the code coverage? There are already parts of the code that we can't fully test e.g: google Maps & sign-in.

app/src/androidTest/java/com/github/h3lp3rs/h3lp/LocalEmergencyCallTest.kt
```diff
@@ -0,0 +1,112 @@
...
+                .getUserCountry(longitude, latitude, targetContext)
+                ?.lowercase()
+            assertEquals(country, actualCountry)
+        }
+    }
+
+    @Test
+    fun getUserCountryWorksOnFalseCoordinates() {
+        assertNull(
+            LocalEmergencyCaller.getUserCountry(
```
AlSchlo, line 88: You can import LocalEmergencyCaller (for all calls). This would make the code more readable.

app/src/main/java/com/github/h3lp3rs/h3lp/LocalEmergencyCaller.kt
```diff
@@ -0,0 +1,105 @@
...
+import android.location.Geocoder
+import com.opencsv.CSVReader
+import java.io.BufferedReader
+import java.io.InputStreamReader
+
+/**
+ * Class containing methods enabling the Call Emergency Services button to make the number called
+ * match to the emergency services number in the user's current location
+ */
+class LocalEmergencyCaller {
```
AlSchlo, line 13: Why bother with a class if it only has a companion object? Might as well simply define LocalEmergencyCaller  as an object directly.

AlSchlo (approved):
Flabbergasting code! Well done! 
I have to agree with Wiam, the documentation is really a pleasure to read.

app/src/main/java/com/github/h3lp3rs/h3lp/LocalEmergencyCaller.kt
```diff
@@ -0,0 +1,105 @@
...
+    companion object {
+        /**
+         * Maximum results returned by the geocoder, kept at 1 since we are only searching
+         * for the user's current country and one result is
```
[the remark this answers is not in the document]
Ahayderi: Indeed woups

app/src/androidTest/java/com/github/h3lp3rs/h3lp/LocalEmergencyCallTest.kt
Ahayderi: That is indeed what I wanted to do at first, the problem comes from the fact that both the Geocoder and the reading of the emergency numbers database requires the application context which I can't seem to be able to access from a simple unit test (in the test folder). Consequently, I can't call the methods correctly to test them from the test folder. 
I found no solution online to get the context from the tests in the test folder (which I think makes sense, a unit test should not require access to the application context).

app/src/main/java/com/github/h3lp3rs/h3lp/LocalEmergencyCaller.kt
Ahayderi: Great remark, it should obviously be an object.

PR #101, opened 2022-03-24
94 better ux for map activities

emduc:
Addresses #94 and implements a better map experience:

-  Adds buttons to choose which utilities to display. Requests are made only once and then cached. 
-  Permissions for localisation is now requested on mainpage and the app cannot be accessed without them.
-  Introduces a map id: the map should now have custom colors which can be modified at any time in the cloud. Unfortunately it does not work at the moment for unknown reasons.


emduc:
I am experimenting on the tests but the code is ready for review. They all pass locally even without any API key so it should be a matter of launching them enough times :)

app/src/androidTest/java/com/github/h3lp3rs/h3lp/NearbyUtilitiesActivityTest.kt
```diff
@@ -42,19 +41,53 @@ class NearbyUtilitiesActivityTest {
...
+
+            // Select defibrillators
+            onView(ViewMatchers.withId(R.id.show_defibrillators_button))
+                .check(matches(isDisplayed()))
+                .perform(click())
+
+        }
+    }
+
     private fun canLaunchMap(intent: Intent) {
```
Lindt19, line 83: Nice modularisation!

app/src/main/java/com/github/h3lp3rs/h3lp/MainPageActivity.kt
```diff
@@ -176,6 +187,38 @@ class MainPageActivity : AppCompatActivity() {
...
+            // Permission was denied.
+            // Display the missing permission error dialog when the fragments resume.
+            locPermissionDenied = true
+        }
+    }
+
+    override fun onResumeFragments() {
+        super.onResumeFragments()
+        if (locPermissionDenied) {
+            Toast.makeText(this, "Permission denied", Toast.LENGTH_SHORT).show()
```
Lindt19, line 210: Be careful with using Toasts, I am not sure they work fine with Cirrus. Maybe check this in case of bugs.

Lindt19, line 210: Maybe you can extract the Toast messages in the strings file

app/src/main/java/com/github/h3lp3rs/h3lp/NearbyUtilitiesActivity.kt
```diff
@@ -101,57 +188,44 @@ class NearbyUtilitiesActivity : AppCompatActivity(), OnMapReadyCallback,
 
                 map.moveCamera(CameraUpdateFactory.newLatLngZoom(myPosition, DEFAULT_MAP_ZOOM))
 
-                if (utility != null) {
-                    findNearbyUtilities()
+                //requestedUtility?.let { findNearbyUtilities(it) }
```
Lindt19, line 191: Maybe you forgot this commented code

Lindt19 (approved):
Very nice Implementation! Good job, the map is stunning 👍🏻

app/src/main/java/com/github/h3lp3rs/h3lp/MainPageActivity.kt
emduc: Oh good to know thanks ! In this case it shouldn't be a problem since the tests run with permission, but I will look carefully into it

app/src/androidTest/java/com/github/h3lp3rs/h3lp/NearbyUtilitiesActivityTest.kt
```diff
@@ -42,19 +41,52 @@ class NearbyUtilitiesActivityTest {
...
+            // Select pharmacies
+            onView(ViewMatchers.withId(R.id.show_pharmacy_button))
+                //.check(matches(isDisplayed()))
+                .perform(click())
+
+            // Select defibrillators
+            onView(ViewMatchers.withId(R.id.show_defibrillators_button))
+                //.check(matches(isDisplayed()))
+                .perform(click())
+
```
Adrinels, line 78: why did you comment the check here?

app/src/main/java/com/github/h3lp3rs/h3lp/NearbyUtilitiesActivity.kt
```diff
@@ -183,17 +255,15 @@ class NearbyUtilitiesActivity : AppCompatActivity(), OnMapReadyCallback,
...
+    private fun showPlaces(places: List<GooglePlace>, utility: String) {
         // Create new markers
         for (place in places) run {
             if (place.containsKey("lat")
                 && place.containsKey("lng")
                 && place.containsKey("name")
                 && place["lat"] != null
-                && place["lng"] != null){
+                && place["lng"] != null
+            ) {
```
Adrinels, line 266: maybe you can store this string as class constant

Adrinels (approved):
Good code !
Nothing more to say. Maybe would need more test later on.

app/src/androidTest/java/com/github/h3lp3rs/h3lp/NearbyUtilitiesActivityTest.kt
emduc: that was to investigate the failed tests on cirrus :)

PR #114, opened 2022-03-26
102 add UI for my skills form

Adrinels:
Add form conntaining all skills a helper can provide in case of emergency.
This skills are store in preferences in a new data Class Skills ready to be used for communication


app/src/androidTest/java/com/github/h3lp3rs/h3lp/MySkillsActivityTest.kt
```diff
@@ -0,0 +1,58 @@
...
+
+@RunWith(AndroidJUnit4::class)
+class MySkillsActivityTest {
+    private val ctx: Context = ApplicationProvider.getApplicationContext()
+
+    @get:Rule
+    val testRule = ActivityScenarioRule(
+        MySkillsActivity::class.java
+    )
+
```
Lindt19, line 30: Just a detail, here you can import Intents and Espresso for example to avoid calling them each time

app/src/main/java/com/github/h3lp3rs/h3lp/MySkillsActivity.kt
```diff
@@ -0,0 +1,84 @@
...
+    override fun onStop() {
+        super.onStop()
+        saveData()
+    }
+
+    fun backHome( view: View){
+        val intent = Intent(this, MainPageActivity::class.java)
+        startActivity(intent)
+    }
+    fun helpDialogue( view: View){
```
Lindt19, line 33: Maybe add a small documention of what this method does

Lindt19 (approved):
Good code 👍🏻 the form is so pleasant to use !

app/src/androidTest/java/com/github/h3lp3rs/h3lp/MySkillsActivityTest.kt
wiam8: Yep, same thing for Espresso

app/src/main/java/com/github/h3lp3rs/h3lp/MySkillsActivity.kt
```diff
@@ -0,0 +1,84 @@
...
+
+    @RequiresApi(Build.VERSION_CODES.O)
+    override fun onStop() {
+        super.onStop()
+        saveData()
+    }
+
+    fun backHome( view: View){
+        val intent = Intent(this, MainPageActivity::class.java)
+        startActivity(intent)
```
wiam8, line 31: The use of startActivity will restart the activity from scratch which is not very efficient. We do not want the onCreate method to be called everytime, we should make use of the onResume method as well. I haven't looked at this in details, but this is an interesting post: https://stackoverflow.com/questions/12408719/resume-activity-in-android

wiam8 (approved):
Great work! You included everything needed in the medical card. 
I would however add the missing documentation for the public methods in the activity: backHome and helpDialogue.

app/src/main/java/com/github/h3lp3rs/h3lp/MySkillsActivity.kt
Adrinels: I will change that.

Thanks for the link !

PR #115, opened 2022-03-26
104 fix medical card form bug

Adrinels:
Fix the MedicalCard empty form bug

When deleting hold entry in the form the activity crash because the app was trying to convert null value to Int

This was solved by adding a new error message when a field is empty


Lindt19 (approved):
All good! Thankfully    there was not so much to change to fix this bug 👍🏻

wiam8 (approved):
Nice catch!

PR #116, opened 2022-03-27
112 add app guide

wiam8:
This PR sets up a guide of the app through the use of tap target prompts, which allow the user to get an introduction to the different features our app has. 

The coverage is low for two reasons: 
*  It is not possible to interact with the target prompts through Espresso.
* I had written a test that checks whether the entire guide is displayed in order to have better coverage. It indeed gives 100% coverage on the Jacoco report. However, it failed on Cirrus 12/14 times. You can check the error log. I tried many different fixes as you can see from my commits ( forcing the screen to stay on while testing by setting the debug flag to FLAG_KEEP_SCREEN_ON, using scenarios, using the onClick function, using findObject of the device.. )  all worked locally but kept failing on Cirrus. 

Screenshot 2022-03-31 at 1 10 09 PM

NB: I fixed a few variables' names according to our naming convention.


emduc (approved):
Nice code, clean and with a lot of nice comments. The style refactoring is also welcome :)

app/src/main/java/com/github/h3lp3rs/h3lp/MainPageActivity.kt
```diff
@@ -72,6 +102,100 @@ class MainPageActivity : AppCompatActivity(), OnRequestPermissionsResultCallback
...
+    }
+
+    /**
+     * Starts the application guide. Uses preferences to only start the guide when the app is
+     * launched for the first time.
+     */
+    private fun startAppGuide() {
+        val prefManager = PreferenceManager.getDefaultSharedPreferences(this)
+        if (!prefManager.getBoolean("didShowGuide", false)) {
+            val prefEditor = prefManager.edit()
```
Ahayderi, line 123: Depending on the merge order, this will have to be modified in accordance with the new local storage

app/src/androidTest/java/com/github/h3lp3rs/h3lp/GuideTest.kt
```diff
@@ -0,0 +1,77 @@
...
+class GuideTest {
+    @get:Rule
+    val testRule = ActivityScenarioRule(
+        MainPageActivity::class.java
+    )
+
+    private val targetContext : Context = ApplicationProvider.getApplicationContext()
+
+    private fun clearPreferences() {
+        val sharedPrefs = PreferenceManager.getDefaultSharedPreferences(targetContext)
```
Ahayderi, line 33: Depending on the merge order, the use of the PreferenceManager might have to be switched to call the local storage (though this should be a simple fix)

Ahayderi (approved):
This is great code and documentation, a pleasure to read. Also, thanks for uniformising some of the ids with our name convention.
I'm sorry that Cirrus wasn't in the mood to pass your tests :/

PR #117, opened 2022-03-29
108 fix firebase double autoconversion bug

Adrinels:
Fix Firebase bug due to Firebase inconsistency.

While using Firebase with double could lead to error. Storing round number like 3.0 as double in db, firebase optimize it as storing as 3 a long. But when retrieving this data it cause a type cast error because the incapacity to cast long to double.

I tried many fix and I ended up with these, using super class Number of kotlin

### DIFF
--- a/app/src/main/java/com/github/h3lp3rs/h3lp/database/FireDatabase.kt
+++ b/app/src/main/java/com/github/h3lp3rs/h3lp/database/FireDatabase.kt
@@ -78,10 +78,11 @@ class FireDatabase(path: String) : Database {
      * @return Future of double
      */
     override fun getDouble(key: String): CompletableFuture<Double> {
-        //This Fix is due to an misconception in firebase:
-        //Storing 3.0 in fire base will automatically transform it in an long integer.
-        //This result when getting it as an type error since long cannot be direct cast as double
-        //This is fix getting the field as the super class number and then casting it with its function
+        // This Fix is due to a misconception in firebase:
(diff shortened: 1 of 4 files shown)


Lindt19 (approved):
Nice strategy ! Good that it fixed the bug 👌🏻

app/src/main/java/com/github/h3lp3rs/h3lp/database/FireDatabase.kt
```diff
@@ -78,7 +78,12 @@ class FireDatabase(path: String) : Database {
      * @return Future of double
      */
     override fun getDouble(key: String): CompletableFuture<Double> {
-        return get(key)
+        //This Fix is due to an misconception in firebase:
+        //Storing 3.0 in fire base will automatically transform it in an long integer.
+        //This result when getting it as an type error since long cannot be direct cast as double
+        //This is fix getting the field as the super class number and then casting it with its function
```
wiam8, line 84: I would just add a space at the beginning of each line. I also fixed a few typos. Feel free to change it again :) 
```suggestion
        // This Fix is due to a misconception in firebase:
        // Storing 3.0 in firebase will automatically transform it into a long integer.
        // This causes a type error when getting it since long cannot be directly cast to double.
        // This is fixed by getting the field as the superclass number and then casting it with its function.

```

wiam8 (approved):
Good thing you were able to catch this early, and thank you for taking care of it :).

PR #118, opened 2022-03-29
Implement mocking database and related tests

AlSchlo:
According to #106, this PR contains the addition of a mocked database that can be used for testing purposes.

This implementation offers the advantage of being able to test usual production code without having to modify it.

Indeed, when writing tests, the developer just has to replace the corresponding databases in the Databases enumeration with MockDatabases they created for this purpose.


Adrinels (approved):
Nice code!
Really good abstraction and test classes.

Lindt19 (approved):
All good ! Very nice abstraction 👍🏻

PR #119, opened 2022-03-30
109 add aed position

Lindt19:
In this task, I completed the map with the positions of defibrillators. I added a new file that contains the coordinates of defibrillators in Lausanne which is used in the map activity.
#109


emduc (approved):
Not much to say, it follows nicely the structure that's already in place for the hospitals and pharmacies. It's a shame that we can't get a more general DB

PR #120, opened 2022-03-30
107 store remote preferences on firebase

emduc:
Addresses #107 and connects the local user preferences to the remote firebase Database. 

The abstraction can be used as is without worrying about the connection to the remote database, remote data is used when available and will have to be pushed later on when the app is closed.

/!\ Warning: In order to have full coverage we would need to have a mock for the authenticated user which we currently retrieve using a FireBase instance


app/src/main/java/com/github/h3lp3rs/h3lp/signin/SignInActivity.kt
```diff
@@ -9,41 +11,53 @@ import androidx.activity.result.contract.ActivityResultContracts
...
 
 class SignInActivity : AppCompatActivity() {
     lateinit var signInClient : SignInInterface<AuthResult>
+    private lateinit var userCookie: LocalStorage
+
+    private fun checkToSAndLaunchIfNotAcceptedElseMain() {
+        // Check ToS agreement
+        userCookie = storageOf(Storages.USER_COOKIE) // Fetch from storage
+        if(!userCookie.getBoolOrDefault(getString(R.string.KEY_USER_AGREE), false)) {
+            val i = Intent(this, PresArrivalActivity::class.java)
```
Ahayderi, line 31: i is not a great variable name ;)

app/src/main/java/com/github/h3lp3rs/h3lp/signin/SignInActivity.kt
```diff
@@ -70,10 +84,24 @@ class SignInActivity : AppCompatActivity() {
...
             }
     }
+
+    /**
+     * Companion used to pass on a global context used to open the local
+     * storages
+     */
+    companion object{
+        @SuppressLint("StaticFieldLeak")
+        lateinit var globalContext: Context
```
Ahayderi, line 98: As discussed, this is not the neatest solution but is quite practical in the context of the local storages (and seems to be recommended on a few sources). We could discuss further to find out if that is the only solution, in which case, a couple of other classes could be simplified by using this global context instead of passing on an activity's context on every call (e.g. LocalEmergencyCall).

app/src/androidTest/java/com/github/h3lp3rs/h3lp/StorageTest.kt
```diff
@@ -0,0 +1,67 @@
...
+class StorageTest {
+
+    // Dummy class for complex types
+    private data class Foo(val a1: Int, val a2: String)
+    // Useful variables
+    private lateinit var stor: LocalStorage
+    private val TEST_KEY = "KEY"
+    private val TEST_SEED = Random(0)
+    private val BYTES_PER_CHAR = 2
+    private val DELTA = 1e-10
```
Ahayderi, line 26: A few of these were common with the Database tests, it might be valuable to put them into a separate utils object/class for testing databases / storages (since we are essentially testing the same functionality)

app/src/androidTest/java/com/github/h3lp3rs/h3lp/StorageTest.kt
```diff
@@ -0,0 +1,67 @@
...
+    @Before
+    fun setup() {
+        globalContext = ApplicationProvider.getApplicationContext()
+        Databases.PREFERENCES.db = MockDatabase()
+        Storages.resetStorage()
+        stor = storageOf(MEDICAL_INFO)
+    }
+
+    @Test
+    fun getAndSetWorksProperly() {
```
Ahayderi, line 37:
```suggestion
    fun getAndSetWorkProperly() {

```

app/src/androidTest/java/com/github/h3lp3rs/h3lp/StorageTest.kt
```diff
@@ -0,0 +1,67 @@
...
+    private val TEST_KEY = "KEY"
+    private val TEST_SEED = Random(0)
+    private val BYTES_PER_CHAR = 2
+    private val DELTA = 1e-10
+
+    @Before
+    fun setup() {
+        globalContext = ApplicationProvider.getApplicationContext()
+        Databases.PREFERENCES.db = MockDatabase()
+        Storages.resetStorage()
```
Ahayderi, line 32: Storages.resetStorage was already imported statically

Ahayderi (approved):
Great clean code, everything was clear! I like the way you handled the local storages, which will make testing through mocking much easier.

I had one small question that remained throughout my reading of the PR, namely, why a LocalStorage isn't a Database, they seem to share most of their methods and to be logically linked. I know that some methods are present in a database that aren't necessary in a local storage (ex: methods related to listeners) and vice-versa (ex: pull) but in my opinion, it would be more logical to have a general Database interface with only getters and setters for Booleans, Strings... then two "lower interfaces", one for a remote database, one for a local database.
Along with a better structure, this might allow us to cut some of the duplication in the tests of the databases and local storages since many tests were only on the getters and setters and thus would only concern the "general database". We could thus merge these tests instead of having 2 nearly identical instances of them.

Obviously, this would require some more work and could be the subject of a separate task, or just a discussion on the subject to see what is more appropriate for the project.

app/src/main/java/com/github/h3lp3rs/h3lp/storage/LocalStorage.kt
```diff
@@ -0,0 +1,134 @@
...
+     */
+    fun pull(){
+        /*if (enableOnlineSync) { // TODO: Need mocked version!
+            // Need to be authenticated if online sync is enabled
+            val uid = FirebaseAuth.getInstance().currentUser!!.uid // TODO: Need mocked version!
+            val db = databaseOf(PREFERENCES)
+            db.getString(uid).exceptionally { JSONObject().toString() }.thenAccept {
+                parseOnlinePrefs(it)
+            }
+        }*/
```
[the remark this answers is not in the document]
AlSchlo: Yes, a simple path + uid as key will solve the issue

PR #121, opened 2022-03-30
111 add call location information

Ahayderi:
This PR abstracts away the code used to access the location manager, thus enabling further testing of the HelpParametersActivity through mocking. With this new location manager, the user's current coordinates are displayed on-screen before they call emergency services (see #111) to allow them to give precise information and, hopefully, to be found earlier.


Adrinels (approved):
Nice Code !
Really good use of  design Pattern !

Nice you thought about the case when user dont allow localisation or no service is avaible

app/src/androidTest/java/com/github/h3lp3rs/h3lp/HelpParametersActivityTest.kt
```diff
@@ -1,106 +1,218 @@
...
+
+    @get:Rule
+    var mRuntimePermissionRule: GrantPermissionRule =
+        GrantPermissionRule.grant(Manifest.permission.ACCESS_FINE_LOCATION)
+
+    @Before
+    fun setUp() {
+        init()
+        clearAllPreferences(ApplicationProvider.getApplicationContext())
+        Preferences(PRESENTATION, ApplicationProvider.getApplicationContext()).setBool(
```
AlSchlo, line 61: You can change this and use our wonderful new local storage class (depending on who merges first)

AlSchlo: Setting the static context variable will be needed as well (see Sign in/Irrelevant presentation tests)

app/src/main/java/com/github/h3lp3rs/h3lp/locationmanager/GeneralLocationManager.kt
```diff
@@ -0,0 +1,32 @@
+package com.github.h3lp3rs.h3lp.locationmanager
+
+object GeneralLocationManager {
+    // locationManager contains the currently used location manager
+    private var locationManager: LocationManagerInterface? = null
```
AlSchlo, line 5: Maybe use lateint?

AlSchlo (approved):
Very nice as usual.
The most important remark is probably about the way to properly create a mock test environment setup (local storage & remote storage). Right now, it's fine, but you'll see how it changes with new syntax & calls [mock authenticated user (later), mock remote database, reset local storage, setup static global context].

app/src/main/java/com/github/h3lp3rs/h3lp/locationmanager/GeneralLocationManager.kt
Ahayderi: In a normal use case locationManager isn't necessarily set before get is called so I'm don't think a lateinit would work here (without checks on initialisation which would defeat its purpose).

AlSchlo (approved):
Good riddance Guava!
We will not miss you!

PR #125, opened 2022-03-31
113 add shortest path map

Lindt19:
For #113 @Ahayderi and I implemented the display of the shortest path to a certain destination on the map.  We included this functionality in the nearby utilities' map. The destination coordinates are fixed for now and can be replaced by the helper's position in the future.


app/src/main/java/com/github/h3lp3rs/h3lp/NearbyUtilitiesActivity.kt
```diff
@@ -297,6 +323,66 @@ class NearbyUtilitiesActivity : AppCompatActivity(), OnMapReadyCallback,
...
+    override fun onPolylineClick(polyline: Polyline) {
+        // Flip from solid stroke to dotted stroke pattern.
+        if (polyline.pattern == null || !polyline.pattern!!.contains(DOT)) {
+            polyline.pattern = PATTERN_POLYLINE_DOTTED
+        } else {
+            // The default pattern is a solid stroke.
+            polyline.pattern = null
+        }
+        Toast.makeText(
+            this, "Fastest path to the user in need of your help",
```
emduc, line 381: This text could be saved in the resources :)

emduc (approved):
Nice code. I like the new abstractions and how you nicely treated the inclusion of the API call, it defines the right structure for our next features :). One thing that I was wondering while looking at the code, you hard-coded a position to represent someone in need, but you already have many positions for all the hospitals, pharmacies and defibrillators nearby, it would be cool to have a way to find the itinerary to a place when we select it. Of couse it is not something you can do in a short amount of time this week but it could be an idea for future improvements.

app/src/main/java/com/github/h3lp3rs/h3lp/NearbyUtilitiesActivity.kt
```diff
@@ -3,35 +3,36 @@ package com.github.h3lp3rs.h3lp
...
 import java.io.InputStreamReader
 import java.lang.Double.parseDouble
 import java.net.HttpURLConnection
 import java.net.URL
+import com.github.h3lp3rs.h3lp.util.GPathJSONParser
+import com.github.h3lp3rs.h3lp.util.JSONParserInterface
+
+
 
 typealias GooglePlace = HashMap<String, String>
```
AlSchlo, line 37: Wow, super cool

AlSchlo (approved):
Very fancy code!
Google API doesn't seem very friendly.

PR #140, opened 2022-04-04
129 update UI

wiam8:
This PR handles: 
- Removing the default action bar and replacing it with our own toolbar that matches our color palette with a few menu items.
- Making the design of the app more cohesive: namely changes to the main page and the help parameters activity.
- Adding meaningful text to the app presentation. 
- Writing tests for the sidebar activity and toolbar to improve our overall coverage.


app/src/main/res/layout/activity_main_page.xml
```diff
@@ -379,6 +387,7 @@
             android:contentDescription="@string/help_button"
             android:onClick="goToHelpParametersActivity" />
```
Adrinels, line 380: This Button seems not to have any vertical layout.

I think we can make it bigger and give it a marginBottom.

This way it will be easier on case of emergency.

app/src/main/res/layout/activity_help_parameters.xml
```diff
@@ -155,7 +175,7 @@
         android:layout_height="wrap_content"
         android:layout_marginStart="44dp"
```
Adrinels, line 156: I think we can keep the phone on red. 
Since it is the most important action to do un case of emergency it should pop out.

app/src/main/res/layout/activity_help_parameters.xml
```diff
@@ -20,10 +20,13 @@
         android:layout_marginTop="2dp"
         android:layout_marginEnd="2dp"
         android:layout_marginBottom="60dp"
-        android:background="@drawable/select_meds_button"
         android:checked="false"
         android:filterTouchesWhenObscured="false"
         android:textOff="@string/first_aid_kit"
+        android:background="@drawable/first_aid_kit"
```
Adrinels, line 26: There is no more feedback on which medication is selected. 

It would be cool that the selected one could be enligthed maybe

Adrinels (approved):
Nice uniformisation the app realy needed it.

I recommand you some improvement but if you dont have time it doesn't matter and can be done in a futur design sprint

app/src/main/res/layout/activity_help_parameters.xml
wiam8: It was initially in green actually haha, but sure :)

app/src/main/res/layout/activity_help_parameters.xml
wiam8: You're indeed correct, I didn't pay attention to this issue. However, I couldn't find a way to highlight the button, the only relevant methods I found were changing the background or text color on selection, both of which are not really a solution. I changed the text to "Selected" once the button is selected and changed the activity's code accordingly.

app/src/androidTest/java/com/github/h3lp3rs/h3lp/ToolbarTest.kt
```diff
@@ -0,0 +1,100 @@
...
+    @get:Rule
+    val testRule = ActivityScenarioRule(
+        MainPageActivity::class.java
+    )
+
+    @Before
+    fun setup() {
+        Intents.init()
+        val intent = Intent()
+        val intentResult = Instrumentation.ActivityResult(Activity.RESULT_OK, intent)
```
Lindt19, line 42: I think there is no need to set the intent result to OK since you only check that it was fired

app/src/main/java/com/github/h3lp3rs/h3lp/MainPageActivity.kt
```diff
@@ -40,8 +41,8 @@ const val HOSPITALS = "Hospitals"
 const val PHARMACIES = "Pharmacies"
```
Lindt19, line 40: You can also add this in the strings.xml

Lindt19 (approved):
Very good code and beautiful UI good job 👍🏻 the app looks stunning 🤩

PR #141, opened 2022-04-06
128 add notification services

Adrinels:
Add NotificationService contiaining usefull method to easily send notification stylised with H3LP app style


app/src/main/java/com/github/h3lp3rs/h3lp/notification/NotificationService.kt
```diff
@@ -0,0 +1,123 @@
...
+
+    companion object {
+        private const val CHANNEL_ID = "Help_channel_id"
+        private var notificationId= 0
+
+
+
+        /**
+         * Create the notification channel for HELP application
+         * Should be called before sending any notification
```
emduc, line 31: Since it should always be called in order to initialize the channel, it would make sense to find a way to do that automatically, either by putting it in the companion object so that it's called once when the app is launched, or in the constructor so that we call it every time we create an instance of NotificationService

Adrinels: That's true.
I was thinking that this will be called when constructing the part that will be running in backgroung of our app reponsible for emergency notification

app/src/main/java/com/github/h3lp3rs/h3lp/notification/NotificationService.kt
```diff
@@ -0,0 +1,121 @@
...
+import android.graphics.BitmapFactory
+import android.os.Build
+import androidx.core.app.NotificationCompat
+import androidx.core.app.NotificationManagerCompat
+import com.github.h3lp3rs.h3lp.R
+
+/**
+ *  Static class containing useful function to throw notification.
+ *  The notification are stylized with H3LP app style
+ *  @warning allways make sure to use {@link createNotificationChannel} method to make sure the notification channel is initialized
```
Ahayderi, line 19: Maybe just add a line return, this comment is a bit longer than the usual limit for line size

app/src/main/java/com/github/h3lp3rs/h3lp/notification/NotificationService.kt
```diff
@@ -0,0 +1,121 @@
...
+         * Send notification in with the given context
+         * Take care of assigning a unique id to the notification
+         * @param notification the notification to send
+         * @param ctx the app context from whem to throw notification
+         */
+        private fun sendNotification(notification : Notification, ctx: Context){
+            with(NotificationManagerCompat.from(ctx)) {
+                notify(notificationId, notification)
+            }
+            notificationId++
```
Ahayderi, line 97: You might want to make the increment atomic (just wrapping it into a synchronized block) since a same NotificationService could possibly be used concurrently

app/src/main/java/com/github/h3lp3rs/h3lp/notification/NotificationService.kt
```diff
@@ -0,0 +1,121 @@
...
+            val bigImageBitmap = BitmapFactory.decodeResource(
+                ctx.resources,
+                R.drawable.notification_icon
+            )
+            return NotificationCompat.Builder(ctx, CHANNEL_ID)
+                .setSmallIcon(R.drawable.notification_icon2)
+                .setLargeIcon(bigImageBitmap)
+                .setContentTitle(title)
+                .setContentText(description)
+                .setColor(0xFF0000)
```
Ahayderi, line 117: You could store the colour in the colors.xml file

app/src/main/res/drawable/notification_icon.xml
```diff
@@ -1,5 +0,0 @@
-<vector android:height="24dp" android:tint="#FF0000"
```
Ahayderi, line 1: Could also be replaced by the colors.xml value (see comment on NotificationService.kt)

app/src/androidTest/java/com/github/h3lp3rs/h3lp/NotificationTest.kt
```diff
@@ -0,0 +1,75 @@
...
+
+    @get:Rule
+    val testRule = ActivityScenarioRule(
+        MainPageActivity::class.java
+    )
+
+
+    @Test
+    fun sendSimpleNotificationWork() {
+        NotificationService.createNotificationChannel(ctx)
```
Ahayderi, line 32: Could be in a startup method (annoted with @Begin) since it's used in both tests though this would not be necessary if you implement Emilien's comments regarding the fact that sending a notification should already create the notification channel if it wasn't already created.

Ahayderi (approved):
Nice and clear code, the notifications look very good and the interface to use them will be a pleasure to work with!

The main issue is the one already raised by Emilien.

app/src/main/java/com/github/h3lp3rs/h3lp/notification/NotificationService.kt
Adrinels: After some reflexion here is what i come up with:
- putting it in init block is impossible since it need the context of an app to be started.
- I read this is always better to recreate the chanel before sending a notification. Because sometime android remove it. That not bad for permormance because createNotificationChannel always check if the notififcation channel is already created and doen't recreated it then
- If we still want to initialize it once I think we could call this at the same time than the database listener which will be normally be done next week

PR #143, opened 2022-04-06
126 set up emergency object

### LINKED ISSUE #126: Set up the object to send to the DB when a user asks for help

Lindt19:
To resolve #126 I implemented an interface to communicate with the database in order to save, retrieve and delete the information describing the emergency situation of the person needing help. The implementation is inspired by the Repository design pattern, which can be used in the future to communicate with the database.

### DIFF
--- a/app/src/main/java/com/github/h3lp3rs/h3lp/HelpParametersActivity.kt
+++ b/app/src/main/java/com/github/h3lp3rs/h3lp/HelpParametersActivity.kt
@@ -16,6 +16,7 @@ import androidx.appcompat.app.AppCompatActivity
 import com.github.h3lp3rs.h3lp.database.models.EmergencyInformation
 import com.github.h3lp3rs.h3lp.database.repositories.emergencyInfoRepository
 import com.github.h3lp3rs.h3lp.locationmanager.GeneralLocationManager
+import com.google.gson.Gson
 import java.util.*
 import kotlin.collections.ArrayList
 
@@ -111,6 +112,10 @@ class HelpParametersActivity : AppCompatActivity() {
      */
     private fun sendInfoToDB(){
         if(latitude != null && longitude != null){
+            val medicalInfo = getSharedPreferences(getString(R.string.medical_info_prefs), MODE_PRIVATE).edit()
+            val gson = Gson()
+            val json = gson.toJson(medicalInfo)
+            Log.d("HERE", json)
             val info = EmergencyInformation(latitude = latitude!!, longitude = longitude!!, meds =  meds, time = currentTime)
             emergencyInfoRepository.create(info)
             emergencyInfoRepository.get(info.id).thenAccept {
--- a/app/src/main/java/com/github/h3lp3rs/h3lp/HelpParametersActivity.kt
+++ b/app/src/main/java/com/github/h3lp3rs/h3lp/HelpParametersActivity.kt
@@ -33,7 +33,7 @@ class HelpParametersActivity : AppCompatActivity() {
     private var latitude: Double? = null
     private var longitude: Double? = null
     private var meds: ArrayList<String> = ArrayList()
-    private var currentTime: Date = Calendar.getInstance().time;
+    private val currentTime: Date = Calendar.getInstance().time;
 
     override fun onCreate(savedInstanceState: Bundle?) {
         super.onCreate(savedInstanceState)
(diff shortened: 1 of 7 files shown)


app/src/main/java/com/github/h3lp3rs/h3lp/HelpParametersActivity.kt
```diff
@@ -23,6 +30,10 @@ class HelpParametersActivity : AppCompatActivity() {
     // userLocation contains the user's current coordinates (is initialized to null since we could
     // encounter an error while getting the user's location)
     private var userLocation: Location? = null
+    private var latitude: Double? = null
+    private var longitude: Double? = null
```
wiam8, line 34: Instead of initializing these variables to null, you could use the lateinit modifier

app/src/main/java/com/github/h3lp3rs/h3lp/HelpParametersActivity.kt
```diff
@@ -89,11 +102,24 @@ class HelpParametersActivity : AppCompatActivity() {
...
+     * Stores the emergency information in the database for further use
+     */
+    private fun sendInfoToDB(){
+        if(latitude != null && longitude != null){
+            val medicalInfo = getSharedPreferences(getString(R.string.medical_info_prefs), MODE_PRIVATE).getString(getString(R.string.medical_info_key),"")
+            val gson = Gson()
+            val medicalInfoJson = gson.toJson(medicalInfo)
+            val emergencyInfo = EmergencyInformation(latitude = latitude!!, longitude = longitude!!, meds =  meds, time = currentTime, medicalInfo = medicalInfoJson)
+            emergencyInfoRepository.create(emergencyInfo)
+        }
```
wiam8, line 120: What happens if the latitude or longitude is null? Do we consider that if one of the two is missing that the information should simply not be sent to the database, which does make sense because otherwise, it is just simply not possible to provide this person with help if the person willing to help can't localize them. However, it would be wise to give feedback to the user stating that we were not able to access his location and therefore can't broadcast his request for help. Either send a reminder that he should accept to share his location, as is done in most applications nowadays, if that was the issue, or try to access the location again if the issue was coming from something else. Anyways, this is not something you have to fix, we should discuss this as a group.

app/src/main/java/com/github/h3lp3rs/h3lp/database/models/EmergencyInformation.kt
```diff
@@ -0,0 +1,15 @@
...
+/**
+ * A model describing an emergency information object
+ */
+data class EmergencyInformation(
+    val id: String = UUID.randomUUID().toString(),
+    val latitude: Double,
+    val longitude: Double,
+    val meds: ArrayList<String>,
+    val time: Date,
+    val medicalInfo: String
```
wiam8, line 14: Maybe add a comment here about what the medicalInfo should include.

app/src/main/java/com/github/h3lp3rs/h3lp/database/repositories/Repository.kt
```diff
@@ -0,0 +1,30 @@
...
+     * @return A CompletableFuture wrapping the value
+     */
+    fun get(id: String): CompletableFuture<T>
+
+    /**
+     * Creates an entry in the database with the given value
+     *
+     * @param value The object value to create
+     */
+    fun create(value: T)
```
wiam8, line 22: Why call this create? Shouldn't it be insert?

wiam8 (approved):
Great work! We're getting closer to having a functional app :)

app/src/main/java/com/github/h3lp3rs/h3lp/database/models/EmergencyInformation.kt
emduc: We can probably leave it as is for now, or define another class to represent that => Don't we have one already?

Lindt19: Yes but we save it as a JSON : String in the database, but originally it's of time MedicalInformation

wiam8: I meant to ask her to just add a comment about what she meant by medicalInfo, because I didn't understand at first what she meant by it. Looking at the changes she made now, she intends to send the entire medical card to the database which should indeed be a separate object and not a String as you said.  However, this is an entirely different task, and as we are dealing with the user's sensitive data, we have to first ask the user to choose which information he is ok with sending to the database and thus share with other users ( blood type, allergies, current medication...). This could simply be a couple of toggle buttons in the settings activity and depending on their value we would choose which elements to send to the database.

emduc (approved):
Nice work overall, I just have a hard time understanding your Repository interface, but we can talk about it after the next stand up meeting :). Make sure that you use our new abstraction for the local storage too, apart from that great job !

app/src/main/java/com/github/h3lp3rs/h3lp/database/repositories/Repository.kt
```diff
@@ -0,0 +1,30 @@
+package com.github.h3lp3rs.h3lp.database.repositories
+
+import java.util.concurrent.CompletableFuture
+
+/**
+ * General interface displaying the different methods required to store, retrieve and delete an object of T in the database
+ */
+interface Repository<T>{
```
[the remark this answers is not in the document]
Lindt19: Actually, it's another level of abstraction. You can think of it as an intermediary between the raw database interface and the user. In this particular case manipulating the emergency information object is straightforward but sometimes you might need to manipulate data before storing it which can be done by the repository.

app/src/main/java/com/github/h3lp3rs/h3lp/HelpParametersActivity.kt
```diff
@@ -89,11 +102,24 @@ class HelpParametersActivity : AppCompatActivity() {
...
             startActivity(intent)
         }
     }
 
+    /**
+     * Stores the emergency information in the database for further use
+     */
+    private fun sendInfoToDB(){
+        if(latitude != null && longitude != null){
+            val medicalInfo = getSharedPreferences(getString(R.string.medical_info_prefs), MODE_PRIVATE).getString(getString(R.string.medical_info_key),"")
```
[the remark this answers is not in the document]
Lindt19: Good point!

PR #146, opened 2022-04-07
130 coverage and refactoring

AlSchlo:
The PR handles #130 to increase the coverage of our codebase by testing what was not as well as to refactor some inconsistencies.

The following changes mainly concern:

- Increasing the coverage for the presentation: SwipeListener. This class wasn't tested until now due to Cirrus problems. By instantiating the listener in a dummy activity and manually passing the arguments of the swipe event, I was able to cover most of its code.
- Mock userUid in a static variable at startup (in the same way we did for the application context) to allow an increase of coverage on the LocalStorage class. This way may seem rather unorthodox, but is in reality very practical and clean. Thus it is a good practice (and some sources on internet agree on this point).
- Uniformize the way of handling local storage and adapting its tests. Until now, most developers implemented their own way of storing local data using preferences. I removed these calls and replaced them with much shorter and neater LocalStorage instances. Sometimes this meant creating new instances in the enumeration. A lot of tests also had to be modified to make this change work, since LocalStorage invocation requires a mocked authentication (in the case of an online connected storage) and the static context must also always be set at test setup.
- As a result of the previous point, the tests were modified by removing a global startup rule for the concerned classes. To keep things modular, I often created a custom launch function to wrap the previous test code. The reason this is **NOT** a big problem is because custom rules automatically call the onCreate method **BEFORE** the `@Before` annotation. This sadly does not scale when we want to customize the way an activity is created. Perhaps rules can be defined by users, and we could potentially modularize a lot of code this way (but I have not had the time to look into this). _THIS SHOULD NOT BE DONE WHEN THE ONCREATE METHOD IS NOT PROBLEMATIC (EX: STORAGEOF)._
- Global static imports in tests, fix typos, remove some magic values, fix indentation problems, etc.

_An important note:_ A lot of code could still be refactored, but I consider my changes to be the most urgent ones. Also, I did not want to go into a logic of rewriting too much. Indeed, you know the saying: _"Don't fix what ain't broken!"_. As a consequence, I kindly ask by reviewers **NOT** to comment too much about the code I did not change. Else we could talk about the whole codebase for an entire day :-)

image


app/src/androidTest/java/com/github/h3lp3rs/h3lp/MedicalCardActivityTest.kt
```diff
@@ -1,126 +1,220 @@
...
+
+    private fun start() {
+        init()
+        val intent = Intent()
+        val intentResult = ActivityResult(Activity.RESULT_OK, intent)
+        intending(anyIntent()).respondWith(intentResult)
+    }
+
+    private fun end() {
+        release()
```
Adrinels, line 61: Dont know if i really like this use of release() and init() without Intents.
Eventhough it is smaller I think we lose a bit in our code clarity not knowing what we are initiating / releasing

app/src/androidTest/java/com/github/h3lp3rs/h3lp/StorageTest.kt
```diff
@@ -23,14 +25,15 @@ class StorageTest {
...
-    private val DELTA = 1e-10
 
     @Before
     fun setup() {
         globalContext = ApplicationProvider.getApplicationContext()
-        Databases.PREFERENCES.db = MockDatabase()
+        userUid = USER_TEST_ID
+        PREFERENCES.db = MockDatabase()
         resetStorage()
-        stor = storageOf(MEDICAL_INFO)
```
Adrinels, line 33: Weird variable name "stor* not really clear what it represent maybe can be change to stroage

app/src/androidTest/java/com/github/h3lp3rs/h3lp/signin/NewUserSignInTest.kt
```diff
@@ -42,6 +47,8 @@ class NewUserSignInTest {
     fun setUp() {
```
Adrinels, line 47: Since setup() is really similar in each test files, wouldn't be a good idea to put it in a utils class in test folder?

app/src/main/java/com/github/h3lp3rs/h3lp/MedicalCardActivity.kt
```diff
@@ -94,7 +90,6 @@ class MedicalCardActivity : AppCompatActivity() {
     /**
      * create the field for the birth year with input check
      */
-    @RequiresApi(Build.VERSION_CODES.N)
```
Adrinels, line 97: This was asked by Calendar class. Are you sure it won't cause problem on old api version without the required instruction?

app/src/main/java/com/github/h3lp3rs/h3lp/storage/LocalStorage.kt
```diff
@@ -22,15 +23,14 @@ class LocalStorage(private val path: String, context: Context, private val enabl
      * @throws NullPointerException if the user is not authenticated AND online sync is enabled.
      */
     fun pull(){
-        /*
-        if (enableOnlineSync) { // TODO: Need mocked version!
+        if (enableOnlineSync) {
             // Need to be authenticated if online sync is enabled
```
Adrinels, line 27: Is there a way for the user to choose if he want the synch?

If no it is a good idea to add it to user stories

Adrinels (commented):
Nice Refactor our code really needed it !

Nothing really important to say about it.

Just sometime by importing a full library we gain in code length but often loose in clarity (cf init and release)

app/src/androidTest/java/com/github/h3lp3rs/h3lp/MedicalCardActivityTest.kt
AlSchlo: It was given as piece of advice on the launchpad (week 1). We all know what it means, and the code is boilerplate either way. So I think keeping it as short as possible is cleaner.

app/src/androidTest/java/com/github/h3lp3rs/h3lp/StorageTest.kt
AlSchlo: Yes will change!

app/src/androidTest/java/com/github/h3lp3rs/h3lp/signin/NewUserSignInTest.kt
AlSchlo: Great remark! I thought about it but didn't realize until AFTER all the changes how many times this appeared. Should be changed at another refactoring, I already spent too much time on this one.

app/src/main/java/com/github/h3lp3rs/h3lp/MedicalCardActivity.kt
AlSchlo: No, I tried removing it and it worked :D
You can see the build succeeds!

app/src/main/java/com/github/h3lp3rs/h3lp/storage/LocalStorage.kt
AlSchlo: Yes, you can add this in the enumeration! You cannot forge your own storage instance dynamically if that is your question. And I don't see where this could be useful. (cf. Storages enumeration)
If your question is about choosing whether data is ever put to the database (for privacy or backup constraints) this could indeed be a story for later! Let's add it!

app/src/androidTest/java/com/github/h3lp3rs/h3lp/GuideTest.kt
```diff
@@ -1,54 +1,49 @@
...
-        assertFalse(prefManager.getBoolean("didShowGuide", false))
+    @Before
+    fun setup() {
+        globalContext = getApplicationContext()
+        userUid = USER_TEST_ID
+        PREFERENCES.db = MockDatabase()
+        resetStorage()
     }
 
+    //PR Note: These tests are artificial coverage because Cirrus is not working :p
```
Ahayderi, line 35:
```suggestion
    // PR Note: These tests are artificial coverage because Cirrus is not working :p
```

app/src/androidTest/java/com/github/h3lp3rs/h3lp/GuideTest.kt
```diff
@@ -1,54 +1,49 @@
...
     @Test
     fun checkThatGuideIsLaunched() {
-        val prefManager = PreferenceManager.getDefaultSharedPreferences(targetContext)
-        assertTrue(prefManager.getBoolean("didShowGuide", false))
+        launch().use { assertTrue(storageOf(USER_COOKIE).getBoolOrDefault(GUIDE_KEY, false)) }
     }
 
+    //PR Note: These tests are artificial coverage because Cirrus is not working :p
```
Ahayderi, line 41:
```suggestion
    // PR Note: These tests are artificial coverage because Cirrus is not working :p
```

app/src/androidTest/java/com/github/h3lp3rs/h3lp/MainPageTestActivity.kt
```diff
@@ -26,26 +34,39 @@ import org.junit.runner.RunWith
...
+        return launch(Intent(getApplicationContext(), MainPageActivity::class.java))
+    }
+
+    private fun launchAndDo(action: () -> Unit) {
+        launch().use {
+            start()
+            action()
+            end()
+        }
+    }
```
Ahayderi, line 59: This is a really nice refactoring! It could even be (along with the start and end) put in a utils testing file to be reused since it's also used in the medical card activity test, search bar test, and others and could prove useful elsewhere.

app/src/androidTest/java/com/github/h3lp3rs/h3lp/MedicalCardActivityTest.kt
```diff
@@ -133,76 +227,73 @@ class MedicalCardActivityTest {
...
     @Test
-    fun backButtonWork(){
-        Intents.init()
-        val intent = Intent()
-        Instrumentation.ActivityResult(Activity.RESULT_OK, intent)
-        onView(withId(R.id.medicalInfoBackButton)).perform(click())
-        Intents.intended(
-            Matchers.allOf(
-                IntentMatchers.hasComponent(MainPageActivity::class.java.name),
+    fun backButtonWork() {
```
Ahayderi, line 286:
```suggestion
    fun backButtonWorks() {
```

app/src/androidTest/java/com/github/h3lp3rs/h3lp/MySkillsActivityTest.kt
```diff
@@ -1,58 +1,65 @@
...
-        Instrumentation.ActivityResult(Activity.RESULT_OK, intent)
-        onView(ViewMatchers.withId(R.id.mySkillsBackButton))
-            .perform(ViewActions.click())
-        Intents.intended(
-            Matchers.allOf(
-                IntentMatchers.hasComponent(MainPageActivity::class.java.name),
-            )
-        )
-        Intents.release()
+    fun backButtonWork() {
```
Ahayderi, line 43:
```suggestion
    fun backButtonWorks() {
```

app/src/main/java/com/github/h3lp3rs/h3lp/dataclasses/HelperSkills.kt
```diff
@@ -0,0 +1,4 @@
+package com.github.h3lp3rs.h3lp.dataclasses
+
+data class HelperSkills(val hasEpipen: Boolean, val hasVentolin: Boolean, val hasInsulin: Boolean,
+                        val knowCPR: Boolean, val hasFirstAidKit: Boolean, val isMedicalPro: Boolean)
```
Ahayderi, line 4:
```suggestion
                        val knowsCPR: Boolean, val hasFirstAidKit: Boolean, val isMedicalPro: Boolean)
```

app/src/main/java/com/github/h3lp3rs/h3lp/signin/SignInActivity.kt
```diff
@@ -95,12 +95,20 @@ class SignInActivity : AppCompatActivity() {
...
         fun getGlobalCtx(): Context {
             return globalContext
         }
+
+        /**
+         * Getter on the userUid
+         */
+        fun getUid(): String? {
+            return userUid
+        }
```
Ahayderi, line 112: I am not sure if a getter is necessary here since userId is already public and no transformation actually occurs in the getter.

app/src/androidTest/java/com/github/h3lp3rs/h3lp/GuideTest.kt
```diff
@@ -1,54 +1,49 @@
...
 
-    private val targetContext : Context = ApplicationProvider.getApplicationContext()
-
-    private fun clearPreferences() {
-        val sharedPrefs = PreferenceManager.getDefaultSharedPreferences(targetContext)
-        val editor = sharedPrefs.edit()
-        editor.clear()
-        editor.apply()
+    private fun launch(): ActivityScenario<MainPageActivity> {
+        return launch(Intent(getApplicationContext(), MainPageActivity::class.java))
```
Ahayderi, line 24: Could possibly be refactored out since it's sued in several test files

app/src/androidTest/java/com/github/h3lp3rs/h3lp/presentation/PresArrivalActivityTest.kt
```diff
@@ -11,29 +10,21 @@ import org.junit.Rule
...
 import org.junit.runner.RunWith
 
-
 @RunWith(AndroidJUnit4::class)
 class PresArrivalActivityTest {
     @get:Rule
     val testRule = ActivityScenarioRule(PresArrivalActivity::class.java)
 
+    private fun checkIsDisplayed(id: Int) {
+        onView(withId(id)).check(matches(isDisplayed()))
```
Ahayderi, line 19: Could also be factored out due to its several uses in test files

Ahayderi (approved):
Very nice refactoring, the added consistency is a real pleasure! I know you only had a handful of hours on this task and thus didn't expect groundbreaking refactoring but was still pleasantly surprised at the number of improvements you were able to do.

My main point was about factoring out some of the methods you created for testing (mostly: launchAndDo, launch, start, end, and checkIsDisplayed) since they are repeated several times in the test files.

app/src/androidTest/java/com/github/h3lp3rs/h3lp/MainPageTestActivity.kt
AlSchlo: Definitely agree, I only realized afterwards the number of times I used this. We can name the class TestSetup or something like that. It will contain all required functions to create custom test environments and launch them properly.

app/src/main/java/com/github/h3lp3rs/h3lp/signin/SignInActivity.kt
AlSchlo: You're right, I just copied the context getter 👍 
Should be done for a next refactoring

app/src/androidTest/java/com/github/h3lp3rs/h3lp/presentation/PresArrivalActivityTest.kt
AlSchlo: Yes, like I said before, I think the time is ripe to do a testing utils class.

app/src/androidTest/java/com/github/h3lp3rs/h3lp/GuideTest.kt
AlSchlo: Will delete this, it was just for you guys ;)

app/src/androidTest/java/com/github/h3lp3rs/h3lp/GuideTest.kt
AlSchlo: Same as my other replies

PR #163, opened 2022-04-15
162 fix sign in crash bug

Adrinels:
This short PR fixed two bug discovered last week.

The sigin bug : when lauching multiple time the application with the same account the sigin crashed.

Notification demo bug : the demo code would not show any notification.


app/src/main/java/com/github/h3lp3rs/h3lp/MainPageActivity.kt
```diff
@@ -290,7 +290,8 @@ class MainPageActivity : AppCompatActivity(), OnRequestPermissionsResultCallback
         db.addListener(getString(R.string.ventolin_db_key), String::class.java) {
             if (it.equals(getString(R.string.help))) {
                 db.setString(getString(R.string.ventolin_db_key), getString(R.string.nothing))
-                NotificationService.sendOpenActivityNotification(this,getString(R.string.emergency), getString(R.string.need_help),MainPageActivity::class.java)
+                NotificationService.createNotificationChannel(this)
```
AlSchlo, line 293: You can import this statically, since there cannot be any confusion about the fact that it is the NotificationService (the function name is explicit enough).

AlSchlo (approved):
Thanks for fixing this!

emduc (approved):
Thanks a lot for the fix :)

PR #176, opened 2022-04-18
Fix tests of scroll buttons

emduc:
Addresses #158 and fixes some of the tests that we needed to remove because of Cirrus. There were both localization and timing issues that I fixed using the mock localization and explicit calls to the root. The timing problem came from the fact that the pop-up is only initiated in the on-create and the test did not focus on the right view (~50% of the time) and therefore was sometimes not able to close the pop-up.


emduc:
I had to cherry pick some tests because Cirrus does not like to scroll or click on button in the scroll view (which works locally), the lacking 40% are 3 lines of code and will not impact the overall coverage. I took the opportunity to refactor the class according to the TA review made over Easter.

app/src/androidTest/java/com/github/h3lp3rs/h3lp/AwaitHelpActivityTest.kt
```diff
@@ -103,56 +150,22 @@ class AwaitHelpActivityTest {
...
-//        phoneButton.perform(click())
-//
-//        intended(
-//            Matchers.allOf(
-//                IntentMatchers.hasAction(Intent.ACTION_DIAL)
-//            )
-//        )
-//    }
 
     private fun getIntent(): Intent {
```
AlSchlo, line 170: This function is unused and can be removed

AlSchlo (approved):
Nicely done! I imagine this wasn't very fun to do, but at least now it is done. Congratulations! Our coverage will thank you!

Please check out my most important remark about the cookie storage and its related database. If you use the cookie storage somewhere, its underlying database has to be mocked or you will run into issues. Else if it not used, you can just remove it in the setup code.

Adrinels (approved):
Clean new test and good fix for old ones.
Nothing to say more than Alexis comment

emduc:
closed through #184

PR #185, opened 2022-04-27
177 180 implement help alert system

AlSchlo:
This PR addresses #180 , #179 , #178 and  #177 .

The full end-to-end alert notification service is thus fully operational. Here are some important points:

**1.** The helper location is not yet displayed on the helpee's screen although present in the object. Indeed, there is a bug in the related activity which makes the map fragment unusable for marking locations. This must be done in another sprint, as it is solely a map related issue.
**2.** The mocked database had an urgent bug which I had to fix in order to emulate the alert systems in tests. The listener had a problem when dealing with complex object types in its callback. This has been solved by distinguishing between native objects (Int, String, Double, Boolean) and the other objects.
**3.** Contains some refactoring of the database to allow more fine-grained listener operations.
**4.** Fixed a critical notification bug that did not allow to link a bundle with an intent.
**5.** Adapted emergency dataclasses, AwaitHelp, HelpParam and HelpPage to make them compatible and consistent.
**6.** Contains 3 important tests that check:
* The helper receives a notification if help is needed and he qualifies 
--> (**Note**: The part where he opens the notifications fails on Cirrus ONLY since there seems to be a dependency between the other notification tests that fail after. This must be explored at a next sprint, and explains why the coverage is not optimal. 
*  The helpee gets notified when a helper (or more) are coming
* The helper notifies the helpee when he accepts help
And some other minor tests such as emergency rejections.
=> These tests are quite big since they require a lot of setup. More tests may be written later on to get a robust implementation as we get close to our final sprints.

If you made it this far congratulations!

Happy code review!


app/src/androidTest/java/com/github/h3lp3rs/h3lp/AwaitHelpActivityTest.kt
```diff
@@ -168,4 +178,46 @@ class AwaitHelpActivityTest {
...
+        EMERGENCIES.db = emergencyDb
+        // Simulate arrival on await page after calling for help
+        launch<AwaitHelpActivity>(intent).use {
+            // Nobody coming
+            onView(withId(R.id.incomingHelpersNumber)).check(matches(withText("")))
+            // One person is coming
+            val helper1 = Helper(USER_TEST_ID + 1, 2.0, 2.0)
+            val withHelpers = emergency.copy(helpers = ArrayList(listOf(helper1)))
+            emergencyDb.setObject(helpId.toString(), EmergencyInformation::class.java, withHelpers)
+            onView(withId(R.id.incomingHelpersNumber)).check(matches(withText("1 person is coming to help you")))
```
Adrinels, line 212: maybe use string ressource for this then when we changed it the test won't break

app/src/androidTest/java/com/github/h3lp3rs/h3lp/MainPageTestActivity.kt
```diff
@@ -134,4 +147,34 @@ class MainPageTestActivity {
...
+        NEW_EMERGENCIES.db = newEmergenciesDb
+        // Add to storage the skills
+        storageOf(SKILLS).setObject(globalContext.getString(R.string.my_skills_key),
+            HelperSkills::class.java, skills)
+        // To track notifications
+        val uiDevice = UiDevice.getInstance(InstrumentationRegistry.getInstrumentation())
+        launchAndDo {
+            // Should immediately receive a notification
+            uiDevice.wait(Until.hasObject(By.textStartsWith("H3LP")),3000)
+            // Get the notification box - CIRRUS DOESN'T LIKE THIS
```
Adrinels, line 173: Weird that it doesn't work anymore seems to work in the notification test last sprint

app/src/main/java/com/github/h3lp3rs/h3lp/HelpPageActivity.kt
```diff
@@ -125,15 +131,44 @@ class HelpPageActivity : AppCompatActivity(), CoroutineScope by MainScope() {
...
+        }
+        databaseOf(EMERGENCIES).getObject(helpId!!, EmergencyInformation::class.java).thenApply {
+            // Add the helper to the list of helpers
+            val me = Helper(userUid!!, currentLat, currentLong)
+            val helpers = ArrayList<Helper>()
+            if(!helpers.contains(me)) {
+                helpers.add(me)
+            }
+            // Stop listening to other emergencies
+            databaseOf(NEW_EMERGENCIES).clearAllListeners()
```
Adrinels, line 164: A discussion about that could be a good idea. 

We could have a case where an helper is closer to another emergency that which he was going and will not be notified

app/src/main/java/com/github/h3lp3rs/h3lp/HelpParametersActivity.kt
```diff
@@ -84,65 +87,128 @@ class HelpParametersActivity : AppCompatActivity() {
...
 
         if (meds.isEmpty()) {
             Toast.makeText(
                 applicationContext,
-                "Please select at least one item", Toast.LENGTH_SHORT
+                getString(R.string.AT_LEAST_ONE_ITEM), Toast.LENGTH_SHORT
             ).show()
         } else {
-
             val b = Bundle()
```
Adrinels, line 110: b is not a realy clear name for a variable

app/src/main/java/com/github/h3lp3rs/h3lp/database/Databases.kt
```diff
@@ -14,5 +25,53 @@ enum class Databases {
...
+                            }
+                            // Never see this emergency again later
+                            emergencyStorage.setBoolean(id.toString(), true)
+                            // Send notification only if the associated object still exists
+                            databaseOf(EMERGENCIES).getObject(id.toString(), EmergencyInformation::class.java)
+                                .thenAccept {
+                                    createNotificationChannel(globalContext)
+                                    val intent = Intent(globalContext, HelpPageActivity::class.java)
+                                    // Data to transfer to the help page activity
+                                    val b = Bundle()
```
Adrinels, line 55: same as before about value name b

app/src/main/java/com/github/h3lp3rs/h3lp/database/Databases.kt
```diff
@@ -14,5 +25,53 @@ enum class Databases {
...
+                }
+                // Chain of listener instantiations
+                activateListener(skills.hasVentolin, globalContext.getString(R.string.asthma_med))
+                activateListener(skills.isMedicalPro, globalContext.getString(R.string.med_pro))
+                activateListener(skills.hasEpipen, globalContext.getString(R.string.epipen))
+                activateListener(skills.hasInsulin, globalContext.getString(R.string.Insulin))
+                activateListener(skills.hasFirstAidKit, globalContext.getString(R.string.first_aid_kit))
+                activateListener(skills.knowsCPR, globalContext.getString(R.string.cpr))
+            }
+        }
```
Adrinels, line 75: This a general remark about the help listener.

Doesn't  listener have only the scope of the activity they were launched in? if this is the case this can cause some problem for our notification.

if the  user close the app or relaunch his phone the listener will be dead then and the user will not get any help notification. This could clearly reduce the utility of our app. Nothing to fix this week but if this is the case we need to add this bug to the backlog

app/src/main/java/com/github/h3lp3rs/h3lp/AwaitHelpActivity.kt
```diff
@@ -88,38 +98,42 @@ class AwaitHelpActivity : AppCompatActivity() {
...
+        helpersId.add(uid)
+        showHelperPerson(uid, latitude, longitude)
+
+        runOnUiThread {
+            findViewById<ProgressBar>(R.id.searchProgressBar).visibility = View.GONE
+            findViewById<TextView>(R.id.progressBarText).visibility = View.GONE
+
+            val helpersText = findViewById<TextView>(R.id.incomingHelpersNumber)
+            if (++helpersNumbers > 1) {
+                helpersText.text = String.format( "%d people are coming to help you", helpersNumbers)
```
Adrinels, line 113: This could be done with string value

Adrinels (approved):
That's some really nice works.
Glad to finally see all the pieces of our app put together. Nothing big to said about your code instead of the remark about the listener code.

However, you might want to split your work into several PRs, especially now that we are in a 2 week sprint. You could for example have split your PR into two: database improvement and communication. It will be easier which PR is doing what. And it will be much more digest for the reviewers

app/src/main/java/com/github/h3lp3rs/h3lp/HelpPageActivity.kt
```diff
@@ -24,7 +29,7 @@ const val EXTRA_DESTINATION_LONG = "help_page_destination_long"
 class HelpPageActivity : AppCompatActivity(), CoroutineScope by MainScope() {
     private lateinit var binding: ActivityHelpPageBinding
 
-    //TODO : currently, the destination is hardcoded, this will change with the task allowing
+    // TODO : currently, the destination is hardcoded, this will change with the task allowing
     // nearby helpers to go and help people in need (in which case the destination will be the
     // location of the user in need)
```
wiam8, line 34: I guess you can remove this TODO now and replace the values with null.

app/src/main/java/com/github/h3lp3rs/h3lp/HelpPageActivity.kt
```diff
@@ -125,15 +131,44 @@ class HelpPageActivity : AppCompatActivity(), CoroutineScope by MainScope() {
...
+    fun acceptHelp(view: View) {
+        if(helpId == null) {
+            goToMainPage()
+            return
+        }
+        databaseOf(EMERGENCIES).getObject(helpId!!, EmergencyInformation::class.java).thenApply {
+            // Add the helper to the list of helpers
+            val me = Helper(userUid!!, currentLat, currentLong)
+            val helpers = ArrayList<Helper>()
+            if(!helpers.contains(me)) {
```
wiam8, line 160: I don't see the need for this check

app/src/main/java/com/github/h3lp3rs/h3lp/signin/GoogleSignInAdapter.kt
```diff
@@ -89,4 +89,12 @@ object GoogleSignInAdapter: SignInInterface<AuthResult> {
     override fun isSignedIn(): Boolean {
         return auth.currentUser != null
     }
+
+    /**
+     * Get tje unique id of the user
```
wiam8, line 94:
```suggestion
     * Get the unique id of the user
```

app/src/main/java/com/github/h3lp3rs/h3lp/signin/SignInInterface.kt
```diff
@@ -36,4 +36,10 @@ interface SignInInterface<T> {
      * Signs the current user out
      */
     fun signOut()
+
+    /**
+     * Get tje unique id of the user
+     * @return uid (null if not signed in)
+     */
+    fun getUid(): String?
```
wiam8, line 44: Good idea to include this!

app/src/main/java/com/github/h3lp3rs/h3lp/database/Databases.kt
AlSchlo: Well I tested this and even when we close the app (and browse the web for example) it still triggers.
When we relaunch the phone, the listeners will get automatically re-added if needed (while it has not been stored locally).

So maybe in some cases where we close the phone, it may not trigger ? I don't know, since the MainActivity will not get destroyed either way...

app/src/androidTest/java/com/github/h3lp3rs/h3lp/MainPageTestActivity.kt
AlSchlo: Yes exactly! But it's the combination of both! I think it's maybe because they have the same name and they are not cleared. Is it possible?

app/src/androidTest/java/com/github/h3lp3rs/h3lp/AwaitHelpActivityTest.kt
AlSchlo: Will do, it has not been done in the code but I should have changed it.

app/src/main/java/com/github/h3lp3rs/h3lp/HelpParametersActivity.kt
AlSchlo: That's not my code but indeed :p

app/src/main/res/layout/activity_help_page.xml
```diff
@@ -101,30 +101,30 @@
...
-                    android:onClick="goToMainPage"
-                    app:layout_constraintTop_toTopOf="parent"
                     app:layout_constraintBottom_toBottomOf="parent"
-                    app:layout_constraintStart_toStartOf="parent" />
+                    app:layout_constraintStart_toStartOf="parent"
+                    app:layout_constraintTop_toTopOf="parent" />
 
-            <ImageView
+                <ImageView
+                    android:id="@+id/accept"
```
wiam8, line 117: This needs to be removed when you merge because when the user accepts to help I have changed the listener to redirect the user to the chat activity.

app/src/main/java/com/github/h3lp3rs/h3lp/database/Databases.kt
AlSchlo: Now it is :(

app/src/main/java/com/github/h3lp3rs/h3lp/AwaitHelpActivity.kt
AlSchlo: Indeed will do

app/src/main/res/layout/activity_help_page.xml
```diff
@@ -101,30 +101,30 @@
                 android:layout_marginBottom="20dp">
 
 
-            <ImageView
+                <ImageView
+                    android:id="@+id/reject"
```
wiam8, line 105: Be careful when you merge! Either remove this or remove the listener added in the activity

AlSchlo:
> 

I totally agree, if I had to do it again, I would probably split it in 2.
I just didn't realize how many changes I had to bring to the existing code base until later on!
Anyways, I detailed the PR a lot, so it still should have been pretty clear!
Thank you!

wiam8 (approved):
Great work! 
Thanks for also taking care of some of the inconsistencies in the different activities and classes. 
Excited to finally see a functional alert system!

app/src/androidTest/java/com/github/h3lp3rs/h3lp/AwaitHelpActivityTest.kt
```diff
@@ -168,4 +178,46 @@ class AwaitHelpActivityTest {
...
+            // The same person is coming again, should NOT add a helper to the list
+            emergencyDb.setObject(helpId.toString(), EmergencyInformation::class.java, withHelpers)
+            onView(withId(R.id.incomingHelpersNumber)).check(matches(withText("1 person is coming to help you")))
+            // A second person is coming
+            val helper2 = Helper(USER_TEST_ID + 2, 2.1, 2.1)
+            val withMoreHelpers = emergency.copy(helpers = ArrayList(listOf(helper1, helper2)))
+            emergencyDb.setObject(helpId.toString(), EmergencyInformation::class.java, withMoreHelpers)
+            onView(withId(R.id.incomingHelpersNumber)).check(matches(withText("2 people are coming to help you")))
+        }
```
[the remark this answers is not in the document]
AlSchlo: I know, I did refactor part of this in the HelpPage, but since I only used this setup once it would have been a bit of a waste of time to do well.
The main issue is that creating a solid testing framework for all these situations would take 10h itself. Since I already went overtime this sprint, I didn't have the time to do it. Though I agree with your points...
Sadly, spending 10h on this won't add any new features :(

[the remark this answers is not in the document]
AlSchlo: Also, even if I put these values outside the function, it won't be any clearer. Like I said, the only real change would be to create an entire testing framework so that we can model complex scenarios. This is literally a full time job itself!

app/src/androidTest/java/com/github/h3lp3rs/h3lp/MainPageTestActivity.kt
```diff
@@ -134,4 +147,34 @@ class MainPageTestActivity {
...
+            // val notification = uiDevice.findObject(By.text(globalContext.getString(R.string.emergency)))
+            // notification.click()
+            // notification.clear()
+            // onView(withId(R.id.accept)).check(matches(isDisplayed()))
+        }
```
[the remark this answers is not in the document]
AlSchlo: See reply above

app/src/main/java/com/github/h3lp3rs/h3lp/HelpPageActivity.kt
AlSchlo: Whoosh, major bug here!
This is amazing since it is not catched my test because the reason is another bug in the existing code base...
Some exciting stuff :-)

AlSchlo: It was meant to be:
val helpers = ArrayList<Helper>(it.helpers)

app/src/androidTest/java/com/github/h3lp3rs/h3lp/AwaitHelpActivityTest.kt
```diff
@@ -168,4 +178,46 @@ class AwaitHelpActivityTest {
...
+            emergencyDb.setObject(helpId.toString(), EmergencyInformation::class.java, withHelpers)
+            onView(withId(R.id.incomingHelpersNumber)).check(matches(withText("1 person is coming to help you")))
+            // The same person is coming again, should NOT add a helper to the list
+            emergencyDb.setObject(helpId.toString(), EmergencyInformation::class.java, withHelpers)
+            onView(withId(R.id.incomingHelpersNumber)).check(matches(withText("1 person is coming to help you")))
+            // A second person is coming
+            val helper2 = Helper(USER_TEST_ID + 2, 2.1, 2.1)
+            val withMoreHelpers = emergency.copy(helpers = ArrayList(listOf(helper1, helper2)))
+            emergencyDb.setObject(helpId.toString(), EmergencyInformation::class.java, withMoreHelpers)
+            onView(withId(R.id.incomingHelpersNumber)).check(matches(withText("2 people are coming to help you")))
```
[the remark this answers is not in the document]
AlSchlo: See above

PR #187, opened 2022-04-27
155 implement settings UI

Adrinels:
This PR add the Settings Activity containing tree main field:

- Account : displaying technical info about the account and allowing user to disconnect
- Synchronization :  allowing user to choose which info should be synchronized with the cloud and allowing the user to delete all his online data 
- Assistance : linking to possible bug report and help activity


app/src/main/java/com/github/h3lp3rs/h3lp/SettingsActivity.kt
```diff
@@ -0,0 +1,109 @@
...
+    }
+
+    /**
+     * Check a given checkbox
+     * @param toggle if it must check or not
+     * @param id the id of the checkbox
+     */
+    private fun check(toggle: Boolean, id: Int){
+        findViewById<CheckBox>(id).isChecked = toggle
+    }
```
Ahayderi, line 92: Since the exact same method is used in the SettingsActivity, it may be useful to refactor it (though I guess this isn't the only instance of such a checkbox and this could be part of a greater refactoring task).
Since the 2 methods have a different name in MySkillsActivity and SettingsActivity, you could name them the same to make a future refactoring easier :D

app/src/main/java/com/github/h3lp3rs/h3lp/SettingsActivity.kt
```diff
@@ -0,0 +1,109 @@
...
+import com.github.h3lp3rs.h3lp.signin.GoogleSignInAdapter.getCreationDate
+import com.github.h3lp3rs.h3lp.signin.GoogleSignInAdapter.signOut
+import com.github.h3lp3rs.h3lp.signin.SignInActivity
+import com.github.h3lp3rs.h3lp.signin.SignInActivity.Companion.getUid
+import com.github.h3lp3rs.h3lp.storage.Storages
+import com.github.h3lp3rs.h3lp.storage.Storages.*
+import com.github.h3lp3rs.h3lp.storage.Storages.Companion.storageOf
+import org.apache.commons.collections4.iterators.SkippingIterator
+
+class SettingsActivity : AppCompatActivity() {
```
Ahayderi, line 22: You could run a reformat file since there's a few spaces missing here and there to be consistent with the rest of the files

app/src/main/java/com/github/h3lp3rs/h3lp/SettingsActivity.kt
```diff
@@ -0,0 +1,109 @@
...
+        findViewById<TextView>(R.id.unique_id_text).text= getUid()
+        findViewById<TextView>(R.id.user_since_text).text= getCreationDate()
+
+    }
+
+    /**
+     * When the user leaves the activity
+     */
+    @RequiresApi(Build.VERSION_CODES.O)
+    override fun onStop() {
```
Ahayderi, line 39: I think this should be an onPause, not onStop so that as soon as the user quits this activity, their changes are saved

app/src/main/java/com/github/h3lp3rs/h3lp/signin/GoogleSignInAdapter.kt
```diff
@@ -97,4 +98,17 @@ object GoogleSignInAdapter: SignInInterface<AuthResult> {
...
+    /**
+     *
+     */
+    fun getCreationDate(): String{
+        val timeStamp = auth.currentUser?.metadata?.creationTimestamp
+        return if(timeStamp==null){
+            ""
+        }else{
+            Date(timeStamp).toString()
+        }
```
Ahayderi, line 111: Using the let/run kotlin nullable operators makes it a little prettier but I'm not sure if it makes it less readable so this is only a suggestion
```suggestion
        return timeStamp?.let { Date(it).toString() } ?: run { "" }
```

app/src/androidTest/java/com/github/h3lp3rs/h3lp/SettingsActivityTest.kt
```diff
@@ -0,0 +1,90 @@
...
+import com.github.h3lp3rs.h3lp.signin.SignInActivity
+import com.github.h3lp3rs.h3lp.signin.SignInActivity.Companion.globalContext
+import com.github.h3lp3rs.h3lp.signin.SignInActivity.Companion.userUid
+import com.github.h3lp3rs.h3lp.storage.Storages
+import com.github.h3lp3rs.h3lp.storage.Storages.Companion.resetStorage
+import org.hamcrest.Matchers.*
+import org.junit.Before
+import org.junit.Test
+import org.junit.runner.RunWith
+import java.lang.NullPointerException
```
Ahayderi, line 28: You could run a reformat code on this file, there's a few imports that can be removed and a few line breaks to add/remove

app/src/androidTest/java/com/github/h3lp3rs/h3lp/SettingsActivityTest.kt
```diff
@@ -0,0 +1,90 @@
...
+        }
+    }
+    @Test
+    fun removeSyncButtonWorks() {
+       launch().use {
+           val db = MockDatabase()
+           db.setString("SKILLS/$USER_TEST_ID","hi")
+           db.setString("USER_COOKIE/$USER_TEST_ID","hi")
+           db.setString("MEDICAL_INFO/$USER_TEST_ID","hi")
+           PREFERENCES.db=db
```
Ahayderi, line 78: Just a heads up that you'll be able to use setDatabase(PREFERENCES, db) at merge

app/src/main/java/com/github/h3lp3rs/h3lp/storage/Storages.kt
```diff
@@ -32,9 +36,25 @@ enum class Storages(enableOnlineSync: Boolean) {
...
+     * @param isSyncEnable if the synchronization must be enabled
+     */
+    fun setOnlineSync(isSyncEnable : Boolean){
+        getGlobalCtx().getSharedPreferences("SyncPref", AppCompatActivity.MODE_PRIVATE)
+            .edit().putString(name, isSyncEnable.toString()).apply()
+    }
+    /**
+     * get the Online synchronization for a given storage
+     */
+    fun getOnlineSync():Boolean{
```
AlSchlo, line 55:
```suggestion
    fun getOnlineSync(): Boolean {
```

app/src/main/java/com/github/h3lp3rs/h3lp/storage/Storages.kt
```diff
@@ -32,9 +36,25 @@ enum class Storages(enableOnlineSync: Boolean) {
...
+            for (storage in values()) {
                 storage.ls.clearAll()
             }
         }
     }
+    /**
+     * set the Online synchronization for a given storage
+     * @param isSyncEnable if the synchronization must be enabled
+     */
+    fun setOnlineSync(isSyncEnable : Boolean){
```
AlSchlo, line 48:
```suggestion
    fun setOnlineSync(isSyncEnable: Boolean) {
```

app/src/main/java/com/github/h3lp3rs/h3lp/storage/Storages.kt
```diff
@@ -32,9 +36,25 @@ enum class Storages(enableOnlineSync: Boolean) {
...
+        getGlobalCtx().getSharedPreferences("SyncPref", AppCompatActivity.MODE_PRIVATE)
```
AlSchlo, line 49: I know this may sound silly, but you can use a storage instance for this ;)

AlSchlo (approved):
Very nice! I would say that some data should not be opt-in remote storage (ex: did they accept the terms of services = COOKIES). Indeed, COOKIES are meant to be sent to an external server. Simply removing these from the UI is enough, since the backend code cannot be accessed by the UI.

app/src/main/java/com/github/h3lp3rs/h3lp/SettingsActivity.kt
```diff
@@ -0,0 +1,109 @@
...
+    /**
+     * Function for the logout button to disconnect from account
+     */
+    fun logout(view: View){
+        signOut()
+        val intent = Intent(this, SignInActivity::class.java)
+        startActivity(intent)
+    }
+
+    fun removeSync(view: View){
```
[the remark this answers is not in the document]
Adrinels: That's totally true will change it

app/src/main/java/com/github/h3lp3rs/h3lp/SettingsActivity.kt
```diff
@@ -0,0 +1,109 @@
...
+
+    }
+
+    /**
+     * Show a dialogue with explication on what is the form for
+     */
+    fun helpDialogue(view: View){
+        AlertDialog.Builder(this).setTitle(getString(R.string.my_helper_skills))
+            .setMessage(getString(R.string.help_my_skills)).show()
+    }
```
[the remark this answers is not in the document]
Adrinels: I plan to implement an Help dialogue for settings but did not do it at the end. I forget this there in the process thanks

app/src/main/java/com/github/h3lp3rs/h3lp/signin/GoogleSignInAdapter.kt
Adrinels: I will use it seems a more "Kotlin" way to do it

app/src/main/java/com/github/h3lp3rs/h3lp/storage/Storages.kt
Adrinels: yeah I hesitated but it seems bit like a snake biting its own tail

PR #188, opened 2022-04-28
153 add new logo and improve design

Adrinels:
This PR introduce the new logo of the app.

It also begin some design refactor on :
 - signin activity
 - presentation activity
This is more like a proposition design than the final design we need to discuss if we want totally change our main color from teal to persimmon or just use it in certain part of the map.

I know that reviewing xml design PR is not the most pleasant thing to do so here is some screenshot : 
image
image
image
image


app/src/main/res/layout/activity_presentation_irrelevant.xml
```diff
@@ -113,52 +125,59 @@
...
 
     <CheckBox
         android:id="@+id/pres3_checkBox"
         android:layout_width="138dp"
         android:layout_height="44dp"
         android:contentDescription="@string/tosbox"
         android:text=""
+        android:textColor="@color/light_gray"
+        android:textColorHighlight="@color/persimmon"
+        android:textColorLink="@color/persimmon"
```
Ahayderi, line 174: I'm not sure of the persimmon for that checkbox, it doesn't appear very well on top of the background

app/src/main/res/layout/activity_presentation_relevant.xml
```diff
@@ -4,16 +4,19 @@
     xmlns:tools="http://schemas.android.com/tools"
     android:layout_width="match_parent"
     android:layout_height="match_parent"
-    tools:context=".presentation.PresRelevantActivity">
+    tools:context=".presentation.PresRelevantActivity"
+    android:background="@drawable/background_grradient_mirror"
```
Ahayderi, line 8: Little typo 
```suggestion
    android:background="@drawable/background_gradient_mirror"
```

app/src/main/res/layout/activity_presentation_irrelevant.xml
Ahayderi, line 174: Same comment as for the checkbox

app/src/main/res/layout/activity_presentation_irrelevant.xml
```diff
@@ -102,6 +113,7 @@
         android:layout_height="46dp"
         android:onClick="sendApproval"
         android:text="@string/ok"
+        android:backgroundTint="@color/persimmon"
```
Ahayderi, line 116: Same comment as for the checkbox

Ahayderi (approved):
Your UI skills never cease to disappoint! I really like this new design and am eager to see how the whole app will look like :eyes: 

As a side note, since I can't directly comment on files, I couldn't leave this as a comment but I'm pretty sure that the ic_launcher_round used as an image for the app is a little off center to the right. The inner heart could also be a little smaller to avoid too much visual conflict on the edges.

app/src/main/res/layout/activity_sign_in.xml
```diff
@@ -21,45 +21,67 @@
...
 
         <ImageButton
             android:id="@+id/signInButton"
-            android:layout_width="265dp"
-            android:layout_height="59dp"
+            android:layout_width="wrap_content"
+            android:layout_height="wrap_content"
             android:adjustViewBounds="false"
-            android:background="@drawable/round_corners"
+            android:background="@drawable/persimmonborder"
```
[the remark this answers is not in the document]
Adrinels: I thinks I will just
 change the image of the button to remove the shadow thanks

AlSchlo (approved):
Congratulations! Your design skills are amazing!
Since I could not comment on the image files, here are some of my (personal) thoughts:
- The color gradient goes a bit too much in the purples, mayble it should be that flashy
- The logo is perfect! Well done.
- There seems to be quite a lot of PNGs added, and they all seem to be pretty much the same (ex: different dimensions). Is it possible to have one sample only and resize them in the code? This might decrease quality, but may reduce the total space of our program.
- I think that the salmon/orange color fits better than the actual green one. So refactoring this is a good idea.
- Same remark about the signin button as Alexandre.
- I don't really like the font of H3LP on the presentation screen. Maybe we should find a more fitting font? I don't know though, I might be a bit to picky.

app/src/main/res/layout/activity_sign_in.xml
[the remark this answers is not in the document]
AlSchlo: Ahh yes, it's the shadow indeed ;)

PR #238, opened 2022-05-02
198 fix location manager

Ahayderi:
This PR addresses #198 by fixing the location manager to use the newest API.
Now, we return the location only when it has been computed to avoid crashes in case the system location manager hasn't gotten it yet :D (which was the cause of the crashes before).


wiam8 (approved):
We won't have to open google maps to make this work 🎉 ! Thanks for fixing this :)

app/src/main/java/com/github/h3lp3rs/h3lp/AwaitHelpActivity.kt
```diff
@@ -265,4 +267,11 @@ class AwaitHelpActivity : AppCompatActivity() {
...
         startActivity(Intent(this, MainPageActivity::class.java))
     }
+
+
+    /** Starts the activity by sending intent */
+    private fun goToActivity(ActivityName: Class<*>?) {
+        val intent = Intent(this, ActivityName)
+        startActivity(intent)
+    }
 }
```
emduc, line 277: I'm not sure to understand why you added this function, you don't seem to use it in this file.

emduc (approved):
Nice ! Thanks a lot for fixing this, everything is clear and seems good to me.

PR #240, opened 2022-05-04
207 - adds end to end encryption

emduc:
addresses #207  and implements end to end encryption within the chat. Keys are generated directly using android's keystore system and all messages sent to the database are now encrypted.


emduc:
Some notes:
The implementation is for now quite primitive because the world of key sharing on fire base is quite mysterious and we struggled to find a working prototype. We lose code coverage from the fact that we now use AndroidKeyStore that cannot be used in unit tests, we're left trying to test Conversation purely through ChatUI [...].

app/src/main/java/com/github/h3lp3rs/h3lp/messaging/ChatActivity.kt
```diff
@@ -91,27 +87,18 @@ class ChatActivity : AppCompatActivity() {
         fun onChildRemoved(key : String) {
             // If the key is the conversation Id, that means that the user deleted the current
```
Ahayderi, line 88: Since the check on key == conversationId now occurs Conversation.kt, you could switch this comment over to there

app/src/main/java/com/github/h3lp3rs/h3lp/messaging/Conversation.kt
```diff
@@ -16,26 +28,140 @@ class Conversation(
...
+            val path = conversationId + "/KEYS/" + Messenger.values()[(currentMessenger.ordinal + 1) % 2].name
+            database.getString(path).thenApply {
+                val decodedKey = Base64.decode(it, Base64.DEFAULT)
+                val keySpecs = X509EncodedKeySpec(decodedKey)
+                publicKey = KeyFactory.getInstance("RSA").generatePublic(keySpecs)
+                sendEncryptedMessage(publicKey!!, messageText) //TODO CHANGE THE !!
+            }
+        } else {
+            sendEncryptedMessage(publicKey!!, messageText)
+        }
```
Ahayderi, line 83:
```suggestion
        publicKey?.let { publicKey ->
            sendEncryptedMessage(publicKey, messageText)
        }?.run {
            val path = conversationId + "/KEYS/" + Messenger.values()[(currentMessenger.ordinal + 1) % 2].name
            database.getString(path).thenApply {
                val decodedKey = Base64.decode(it, Base64.DEFAULT)
                val keySpecs = X509EncodedKeySpec(decodedKey)
                publicKey = KeyFactory.getInstance("RSA").generatePublic(keySpecs)
                publicKey ?.let {
                    sendEncryptedMessage(publicKey!!, messageText)
                }
            }
        }
```

app/src/main/java/com/github/h3lp3rs/h3lp/messaging/Conversation.kt
```diff
@@ -1,7 +1,19 @@
...
+import androidx.security.crypto.MasterKey
 import com.github.h3lp3rs.h3lp.database.Databases.Companion.databaseOf
 import com.github.h3lp3rs.h3lp.database.Databases.MESSAGES
+import java.security.*
+import java.security.spec.MGF1ParameterSpec
+import java.security.spec.X509EncodedKeySpec
+import javax.crypto.Cipher
+import javax.crypto.spec.OAEPParameterSpec
+import javax.crypto.spec.PSource
+import kotlin.text.Charsets.UTF_8
```
Ahayderi, line 16: You can run a reformat, there's a few imports too many

app/src/main/java/com/github/h3lp3rs/h3lp/messaging/Conversation.kt
```diff
@@ -16,26 +28,140 @@ class Conversation(
...
+
+        val encryptedMessage = Message(currentMessenger, Base64.encodeToString(bytes, Base64.DEFAULT), ""/*iv.toString(UTF_8)*/)
+
+        allMessages[encryptedMessage.message] = message
+        database.addToObjectsListConcurrently("$conversationId/MSG/", Message::class.java, encryptedMessage)
+    }
     /**
      * Adds a listener on the conversation, the listener is triggered every time a new message is sent to the
-     * conversation
+     * conversation. act on the list of all messages
```
Ahayderi, line 102: there seems to be a typo

app/src/main/java/com/github/h3lp3rs/h3lp/messaging/Conversation.kt
```diff
@@ -16,26 +28,140 @@ class Conversation(
...
+            }
+        } else {
+            sendEncryptedMessage(publicKey!!, messageText)
+        }
     }
 
+    /**
+     *
+     */
+    private fun sendEncryptedMessage(publicKey: PublicKey, message: String) {
```
Ahayderi, line 89: This method name seems to hint to the fact that sendMessage doesn't actually send encrypted messages, maybe changing its name to something like encryptAndSendMessage would be clearer

app/src/androidTest/java/com/github/h3lp3rs/h3lp/ChatUiTest.kt
```diff
@@ -32,28 +42,67 @@ class ChatUiTest {
...
+
+        // Public key that would be on the database
+        val encodedPublicKey = Base64.encodeToString(kp.public.encoded, Base64.DEFAULT)
+
+        db = MockDatabase()
+        Databases.setDatabase(Databases.MESSAGES, db)
+
+        // Mock the public keys
+        db.setString(CONVERSATION_ID + "/KEYS/" + toMessenger.name, MOCK_KEY)
+        db.setString(CONVERSATION_ID + "/KEYS/" + currentMessenger.name, MOCK_KEY)
```
Ahayderi, line 90: If you extract "/KEYS/" into a const val as suggested in a comment below, you can use it here too!

Ahayderi (approved):
Nice PR! This looks like it really was quite complicated to implement so congrats on making it work :clap: 
I've left a few comments but they should be quite fast to tackle.

app/src/androidTest/java/com/github/h3lp3rs/h3lp/ChatUiTest.kt
```diff
@@ -32,28 +42,67 @@ class ChatUiTest {
     private lateinit var conversationFrom: Conversation
     private lateinit var conversationTo: Conversation
 
+    private lateinit var db: Database
```
AlSchlo, line 45: I'm not sure why this is defined as a lateinit variable, indeed, you can simply access db using databaseOf(MESSAGES) once you have established the mocked version in the test setup function. This is to be more consistent with other test classes where it is done this way.

app/src/androidTest/java/com/github/h3lp3rs/h3lp/ChatUiTest.kt
```diff
@@ -32,28 +42,67 @@ class ChatUiTest {
...
+                KeyProperties.KEY_ALGORITHM_RSA,
+                Conversation.ANDROID_KEY_STORE
+            )
+
+        kpg.initialize(
+            KeyGenParameterSpec.Builder(
+                "key", KeyProperties.PURPOSE_ENCRYPT
+                        or KeyProperties.PURPOSE_DECRYPT
+            )
+                .setEncryptionPaddings(KeyProperties.ENCRYPTION_PADDING_RSA_OAEP)
```
AlSchlo, line 73: KeyProperties can be statically imported

app/src/androidTest/java/com/github/h3lp3rs/h3lp/ChatUiTest.kt
```diff
@@ -32,28 +42,67 @@ class ChatUiTest {
...
+                .setDigests(KeyProperties.DIGEST_SHA1)
```
AlSchlo, line 74: BEWARE!
SHA1 has been cracked and is no longer safe, maybe you should use another hash function? (ex: Keccak or SHA3).

app/src/androidTest/java/com/github/h3lp3rs/h3lp/ChatUiTest.kt
```diff
@@ -32,28 +42,67 @@ class ChatUiTest {
...
+
+        val kp = kpg.generateKeyPair()
+        foreignUserPrivateKey = kp.private
+        foreignUserPublicKey = kp.public
+
+        // Public key that would be on the database
+        val encodedPublicKey = Base64.encodeToString(kp.public.encoded, Base64.DEFAULT)
+
+        db = MockDatabase()
+        Databases.setDatabase(Databases.MESSAGES, db)
```
AlSchlo, line 86: Databases can be imported statically.

app/src/androidTest/java/com/github/h3lp3rs/h3lp/ChatUiTest.kt
```diff
@@ -32,28 +42,67 @@ class ChatUiTest {
...
+
+        // Mock the public keys
+        db.setString(CONVERSATION_ID + "/KEYS/" + toMessenger.name, MOCK_KEY)
```
AlSchlo, line 89: Magic /KEYS/, you can store a constant with the formatted string (ie: %s) so that you can call a String.format and do it more nicely.

app/src/androidTest/java/com/github/h3lp3rs/h3lp/ChatUiTest.kt
```diff
@@ -32,28 +42,67 @@ class ChatUiTest {
...
+        db.setString(CONVERSATION_ID + "/KEYS/" + currentMessenger.name, MOCK_KEY)
+
+
+
 
         conversationFrom = Conversation(CONVERSATION_ID, currentMessenger)
         conversationTo = Conversation(CONVERSATION_ID,toMessenger)
 
+        ActivityScenario.launch<ChatActivity>(intent)
```
AlSchlo, line 98: You can import ActivityScenario statically (I love static imports like you see).

app/src/androidTest/java/com/github/h3lp3rs/h3lp/ChatUiTest.kt
```diff
@@ -32,28 +42,67 @@ class ChatUiTest {
...
+
         init()
     }
 
     @After
     fun release() {
-        conversationFrom.deleteConversation()
+        //conversationFrom.deleteConversation()
         Intents.release()
```
AlSchlo, line 106: And another one

app/src/androidTest/java/com/github/h3lp3rs/h3lp/ChatUiTest.kt
```diff
@@ -63,13 +112,14 @@ class ChatUiTest {
...
+            //.check(matches(hasDescendant(withText(SENT_MESSAGE))))
     }
+
     @Test
     fun receiveMessageDisplaysTheCorrectMessage() {
         conversationTo.sendMessage(RECEIVED_MESSAGE)
         onView(withId(R.id.recycler_view_chat))
             .perform(RecyclerViewActions.scrollToPosition<ViewHolder>(0))
-            .check(matches(hasDescendant(withText(RECEIVED_MESSAGE))))
+            //.check(matches(hasDescendant(withText(RECEIVED_MESSAGE))))
```
AlSchlo, line 123: This code does not check anything. If this is a Cirrus bug, you can add a comment so that we know it for the future.

app/src/androidTest/java/com/github/h3lp3rs/h3lp/ConversationTest.kt
```diff
@@ -0,0 +1,79 @@
...
+//            .getObject(CONVERSATION_ID, Message::class.java)
+//            .orTimeout(30, TimeUnit.SECONDS)
+//
+//        // Checking that the get object failed and that the failure message contained the key we
+//        // were trying to access
+//        //val exception = assertFailsWith<CompletionException> { future.join() }
+//        //assertThat(exception.message, containsString(CONVERSATION_ID))
+//        assertTrue(future.isCompletedExceptionally)
+//    }
+//}
```
AlSchlo, line 79: Why is this class commented? Maybe you can add a comment to explain it (I guess Cirrus ;))

app/src/main/java/com/github/h3lp3rs/h3lp/messaging/Conversation.kt
```diff
@@ -16,26 +22,153 @@ class Conversation(
     private val currentMessenger: Messenger
 ) {
     private val database = databaseOf(MESSAGES)
+    private var publicKey: PublicKey? = null
+    private val allMessages: HashMap<String, String> = HashMap()
```
AlSchlo, line 26: Beware, this object gets lost when the activity closes. Isn't a storage more preferable?

app/src/main/java/com/github/h3lp3rs/h3lp/messaging/Conversation.kt
```diff
@@ -16,26 +22,153 @@ class Conversation(
...
     /**
      * Sends a message from the current user to the database
      * @param messageText The message text
      */
     fun sendMessage(messageText: String) {
-        val message = Message(currentMessenger, messageText)
-        database.addToObjectsListConcurrently(conversationId, Message::class.java, message)
+        // Retrieve public key from Bob
+        publicKey?.let { publicKey ->
+            encryptAndSend(publicKey!!, messageText)
```
AlSchlo, line 68: You can remove !! since you are in a let

app/src/main/java/com/github/h3lp3rs/h3lp/messaging/Conversation.kt
```diff
@@ -16,26 +22,153 @@ class Conversation(
...
+            val kpg =
+                KeyPairGenerator.getInstance(KeyProperties.KEY_ALGORITHM_RSA, ANDROID_KEY_STORE)
+
+            kpg.initialize(
+                KeyGenParameterSpec.Builder(
+                    keyAlias, KeyProperties.PURPOSE_ENCRYPT
+                            or KeyProperties.PURPOSE_DECRYPT
+                )
+                    .setEncryptionPaddings(KeyProperties.ENCRYPTION_PADDING_RSA_OAEP)
+                    .setDigests(KeyProperties.DIGEST_SHA1)
```
AlSchlo, line 45: CF remark about broken SHA1

app/src/main/java/com/github/h3lp3rs/h3lp/messaging/Conversation.kt
```diff
@@ -16,26 +22,153 @@ class Conversation(
...
+    private fun encryptAndSend(publicKey: PublicKey, message: String) {
+        val cipher = Cipher.getInstance(TRANSFORMATION)
+        cipher.init(Cipher.ENCRYPT_MODE, publicKey)
+
+        val bytes = cipher.doFinal(message.toByteArray())
+
+        val encryptedMessage =
+            Message(currentMessenger, Base64.encodeToString(bytes, Base64.DEFAULT))
+
+        allMessages[encryptedMessage.message] = message
```
AlSchlo, line 97: This mapping get deleted when the activity closes => It should be recomputed every time?
You should definitely put it in a local storage (not connected to internet, obviously ;))

app/src/main/java/com/github/h3lp3rs/h3lp/messaging/Conversation.kt
```diff
@@ -16,26 +22,153 @@ class Conversation(
...
+        publicKey?.let { publicKey ->
+            encryptAndSend(publicKey!!, messageText)
+        }?.run {
+            val path = conversationId + "/" + KEYS_SUB_PATH + "/" +
+                    Messenger.values()[(currentMessenger.ordinal + 1) % 2].name
+
+            database.getString(path).thenApply {
+                val decodedKey = Base64.decode(it, Base64.DEFAULT)
+                val keySpecs = X509EncodedKeySpec(decodedKey)
+                publicKey = KeyFactory.getInstance("RSA").generatePublic(keySpecs)
```
AlSchlo, line 76: RSA should definitely be a constant if we want to change our encryption mode later on

AlSchlo (approved):
Pineapple of technology right there! Amazing stuff!
My remarks are mostly minor, except the fact that the decrypted messages should be stored in the storage to avoid multiple decryptions (cache) and it is also more robust.
Also we have storages as a key->value implementation, so might as well use it ;)
Great work! Not only are our user's life safe with us, but also their data!
We truly safe lives!

app/src/androidTest/java/com/github/h3lp3rs/h3lp/ConversationTest.kt
emduc: As I explained above, this can no longer be a unit test since it uses android's key store

app/src/main/java/com/github/h3lp3rs/h3lp/messaging/Conversation.kt
emduc: Oh boy if only it was that easy to change the encryption mode :)

PR #247, opened 2022-05-12
230 forum structure

Ahayderi:
This PR addresses #230, #231, #215, and #222 to create the entire backend of our Forum.
Though it touches quite a few files, PR isn't that long since it only consists in a few objects that define the Forum, their implementation as well as a cache.
Nevertheless, the implementation was a bit complicated since it faced fundamental problems with the way Firebase's database works but the entire "forum communication" process is explained separately in the file ForumProtocol.md in case you want to understand why we set keys / values where we do.
Note: the modifications on the database are minor, simply adjusting to what was changed in main and changing a return type for convenience in the forum; **the files on which you should concentrate are definitely those in the forum package.**

All in all, I hope you like futures because there's *a few* in here :D


app/src/main/java/com/github/h3lp3rs/h3lp/forum/FireForum.kt
```diff
@@ -0,0 +1,262 @@
...
+            }
+            return future
+        } else {
+            // In case we are in a category, we only have a single CategoryPost to return (thus
+            // we return a singleton list)
+            return getAllFromCategory().thenApply { listOf(it) }
+        }
+    }
+
+    private fun getAllFromCategory(): CompletableFuture<CategoryPosts> {
```
Adrinels, line 107: Eventhough it is private it could have some function description for understanding purposes

Adrinels (approved):
Really Nice code!
The interface is really good documented and seems really pleasant to use. great job.

Maybe just one or two thing to say:

1. Is it a good idea to have redundancy between MedicalType and ForumCategory. Wouldn't that cause some difficulty when linking them together. Maybe we can now remplace medicalType with a list of ForumCategory.

2. It seems that the profil.png image has already changed color from teal to persimmon. (Probably my fault but dont know how it happened). It will be better for the presentation if we have the teal version this week. Just roll back this file and it will be fine.

app/src/main/java/com/github/h3lp3rs/h3lp/forum/FireForum.kt
```diff
@@ -0,0 +1,262 @@
...
+     * Abstracts away the implementation of the path into the Forum structure, here, a category is
+     * represented by a path of length 1
+     * @return A boolean corresponding to the fact that this forum is (or isn't) the root forum
+     */
+    private fun isCategory(): Boolean {
+        return path.size == 1
+    }
+
+    /**
+     * Returns the category of the current forum and the default category "general" in case the
```
Lindt19, line 246: Just for consistency, maybe put this description in the @return tag

app/src/main/java/com/github/h3lp3rs/h3lp/forum/FireForum.kt
```diff
@@ -0,0 +1,262 @@
...
+     * current forum is root
+     */
+    private fun getCurrentCategory(): ForumCategory {
+        return if (!isRoot()) {
+            valueOf(path.first())
+        } else {
+            // The default forum category is "general"
+            GENERAL
```
Lindt19, line 254: Would be nice to put this in a general const variable like DEFAULT_CATEGORY in case it changes in the future and if some other classes need to know the default value for the forum category.

app/src/main/java/com/github/h3lp3rs/h3lp/forum/FireForumPost.kt
```diff
@@ -0,0 +1,25 @@
+package com.github.h3lp3rs.h3lp.forum
+
+import com.github.h3lp3rs.h3lp.forum.data.ForumPostData
+import java.util.concurrent.CompletableFuture
+
+/**
```
Lindt19, line 6: Forgotten comments ?

Lindt19 (approved):

Lindt19 (commented):
Very impressed by how you managed to manipulate cleverly all the Futures in your implementation! Keep up the good work 🔥

app/src/main/java/com/github/h3lp3rs/h3lp/forum/FireForum.kt
Ahayderi: > Really Nice code! The interface is really good documented and seems really pleasant to use. great job.
> 
> Maybe just one or two thing to say:
> 
>     1. Is it a good idea to have redundancy between MedicalType and ForumCategory. Wouldn't that cause some difficulty when linking them together. Maybe we can now remplace medicalType with a list of ForumCategory.
> 
>     2. It seems that the profil.png image has already changed color from teal to persimmon. (Probably my fault but dont know how it happened). It will be better for the presentation if we have the teal version this week. Just roll back this file and it will be fine.

Thanks for the suggestions, they've all been implemented

app/src/main/java/com/github/h3lp3rs/h3lp/forum/FireForumPost.kt
Ahayderi: Forgot to delete the class woupsies

PR #263, opened 2022-05-14
258 implement direct sign in

Lindt19:
To address #258 I implemented the ability to be directly redirected to the main page if the user is already signed in. This is done by using the local storage, and thus, enables having a fully offline mode that does not rely on Firebase.


app/src/androidTest/java/com/github/h3lp3rs/h3lp/signin/GoogleSignInTest.kt
```diff
@@ -56,6 +57,9 @@ class GoogleSignInTest : H3lpAppTest() {
         val signInMock = mock(SignInInterface::class.java)
         When(signInMock.isSignedIn()).thenReturn(false)
 
+        val userSignIn = Storages.storageOf(Storages.SIGN_IN)
```
wiam8, line 60: You can import statically storageOf and SIGN_IN;)

app/src/androidTest/java/com/github/h3lp3rs/h3lp/signin/NewUserSignInTest.kt
```diff
@@ -55,6 +56,9 @@ class NewUserSignInTest : H3lpAppTest() {
         val signInMock = mock(SignInInterface::class.java)
         When(signInMock.isSignedIn()).thenReturn(false)
 
+        val userSignIn = Storages.storageOf(Storages.SIGN_IN)
```
wiam8, line 59: Same comment as above

app/src/main/java/com/github/h3lp3rs/h3lp/signin/SignInActivity.kt
```diff
@@ -38,14 +39,24 @@ class SignInActivity : AppCompatActivity() {
...
 
+    /**
+     * Save the user authentication information to the local storage
+     */
+    private fun saveAuthentication(){
+        userSignIn.setBoolean(getString(R.string.KEY_USER_SIGNED_IN), true)
+        userUid?.let { userSignIn.setString(getString(R.string.KEY_USER_UID), it) }
+        username?.let { userSignIn.setString(getString(R.string.KEY_USER_NAME), it) }
+    }
+
```
wiam8, line 59: You could define all three as variables as they are used twice:
getString(R.string.KEY_USER_SIGNED_IN)
getString(R.string.KEY_USER_UID)
getString(R.string.KEY_USER_NAME)

app/src/main/java/com/github/h3lp3rs/h3lp/storage/LocalStorage.kt
```diff
@@ -29,7 +29,7 @@ class LocalStorage(private val path: String, context: Context) {
     fun pull(){
         if (enableOnlineSync) {
             // Need to be authenticated if online sync is enabled
-            val uid = getUid()!!
+             val uid = getUid()!!
```
wiam8, line 32:
```suggestion
            val uid = getUid()!!
```

app/src/main/java/com/github/h3lp3rs/h3lp/storage/Storages.kt
```diff
@@ -7,11 +7,11 @@ import java.lang.Boolean.parseBoolean
 /**
  * Enumeration of all useful (local) storages in H3LP
  */
-enum class Storages {
-    USER_COOKIE, MEDICAL_INFO, SKILLS, EMERGENCIES_RECEIVED, FORUM_THEMES_NOTIFICATIONS, FORUM_CACHE;
+enum class Storages(private var isFresh: Boolean) {
```
wiam8, line 10: I'm not sure I understand by what you mean by isFresh

app/src/main/res/values/strings.xml
```diff
@@ -279,6 +279,10 @@
     <string name="first_aid_guide_prompt">Basic first aid training tips and procedures for any emergency !</string>
     <string name="guide_search_bar_prompt">In case of emergency, you can quickly find what you\'re looking for thanks to our available search bar! </string>
     <string name="KEY_USER_AGREE">USER_AGREE</string>
+    <string name="KEY_USER_SIGNED_IN">USER_SIGNED_IN</string>
+    <string name="KEY_USER_UID">USER_UID</string>
+    <string name="KEY_USER_NAME">USER_NAME</string>
```
wiam8, line 284: If these are only used in the code, it is maybe better to define them as public constants in one of the files they are being used. This is to avoid having to use to getString(id) method to get them

wiam8 (approved):
This was very much needed. Great job as always :)

app/src/main/java/com/github/h3lp3rs/h3lp/signin/SignInActivity.kt
```diff
@@ -38,14 +39,24 @@ class SignInActivity : AppCompatActivity() {
     /**
      * Check if the current user is already signed in and update activity accordingly
      */
-    private fun checkIfSignedIn() {
-        if (signInClient.isSignedIn()) {
-            userUid = signInClient.getUid()
-            username = getInstance().currentUser?.displayName?.substringBefore(" ")
+    private fun offlineCheckIfSignedIn(){
+        userSignIn = storageOf(Storages.SIGN_IN) // Fetch from storage
```
emduc, line 43: You can import SIGN_IN statically  here too

app/src/main/java/com/github/h3lp3rs/h3lp/signin/SignInActivity.kt
```diff
@@ -57,14 +68,18 @@ class SignInActivity : AppCompatActivity() {
...
+        offlineCheckIfSignedIn()
     }
 
     /**
      * Initialize client and launch the sign in request
      */
     private fun launchSignIn(){
         signInClient = SignIn.get()
-        checkIfSignedIn()
+        //checkIfSignedIn()
```
emduc, line 82: is that still necessary? If not we can just remove it

app/src/main/java/com/github/h3lp3rs/h3lp/signin/SignInActivity.kt
```diff
@@ -86,6 +101,7 @@ class SignInActivity : AppCompatActivity() {
                     userUid = signInClient.getUid()
                     // Only get the first name for privacy reasons
                     username = getInstance().currentUser?.displayName?.substringBefore(" ")
+                    saveAuthentication()
                     checkToSAndLaunchIfNotAcceptedElseMain()
                 }
```
emduc, line 106:
```suggestion
                    username = getInstance().currentUser?.displayName?.substringBefore(" ")
                    
                    saveAuthentication()
                    checkToSAndLaunchIfNotAcceptedElseMain()
                }
```

app/src/main/java/com/github/h3lp3rs/h3lp/storage/Storages.kt
```diff
@@ -7,11 +7,11 @@ import java.lang.Boolean.parseBoolean
...
+    USER_COOKIE(false), MEDICAL_INFO(false), SKILLS(false), EMERGENCIES_RECEIVED(false), FORUM_THEMES_NOTIFICATIONS(false), FORUM_CACHE(false), SIGN_IN(true);
 
     private val ls = LocalStorage(name, getGlobalCtx())
-    private var isFresh = false
+    //private var isFresh = false
```
emduc, line 14: It seems like a massive change, why?

When we implemented the storage, we used this variable so that we always pull the very first time a storage is used and therefore there is no need to know in the code if the data is freshed or not, if the storage has been used somewhere else then it's freshed otherwise it's not. I don't understand what is the problem of pulling in your case (ie. having isFresh false)

emduc (approved):
Nice work, it needed to be done thank you.
I'm a bit worried about your change with the storages, I'd love to know more about why you did it that way :)

app/src/main/java/com/github/h3lp3rs/h3lp/storage/Storages.kt
Lindt19: Well it's not a massive change, all the other storages are kept "not fresh" at the beginning. The thing is that the purpose of using a storage SIGNIN is to allow a completely offline mode with a sign-in option. Which means not using the remote database at all. That's why I also disabled online synchronization. Also, the first call to pull needs a user ID, which we don't have yet since the user is not signed in yet (So it was actually impossible to keep the pull) this is not a problem for the other storages since they are in a context where we already have a user Id. Another reason is that I only store the username and the user Id, and this information is supposed not to change unless another user is signed in, thus, I feel like there is no need for checking if they conform to a remote database.

I am aware it was not the best solution to make isFresh true for the SIGNIN storage, but at this point I feel like it's the simplest way to do it. We can still change that in a future task if the time allows it😉

app/src/androidTest/java/com/github/h3lp3rs/h3lp/signin/GoogleSignInTest.kt
Lindt19: SIGNIN can't be statically imported

app/src/main/java/com/github/h3lp3rs/h3lp/storage/Storages.kt
emduc: Hehe yes thank you for the explanation. After my PR the onlineSync will be false by default (a lot of storages are onlineSynced but shouldn't be) so this won't be necessary, but I can make the change when I merge after you

app/src/androidTest/java/com/github/h3lp3rs/h3lp/signin/GoogleSignInTest.kt
wiam8: use this : import com.github.h3lp3rs.h3lp.storage.Storages.SIGN_IN

app/src/main/res/values/strings.xml
Lindt19: It was defined like this by the teammates who implemented the local storage and it makes more sense to keep them like that because they are keys of the local storage, and thus can be accessed from anywhere in the app

app/src/main/java/com/github/h3lp3rs/h3lp/signin/SignInActivity.kt
Lindt19: Can't be done unfortunately since it's an enum value :(

app/src/main/java/com/github/h3lp3rs/h3lp/signin/SignInActivity.kt
Lindt19: Yes just forgotten oups 😬

app/src/main/java/com/github/h3lp3rs/h3lp/storage/Storages.kt
Lindt19: Oh cool that you did the change on your PR  👍🏻

PR #267, opened 2022-05-16
257 261 fix notifications bug

wiam8:
The bug mentioned in #257 was due to a change in the type of MedicalType which somehow previously allowed having null parameters, which caused a null pointer exception in the local storage after changing the type of parameters as I explained in the last meeting. To fix this you have to reset your local storage.

Previously, when a professional accepted to receive notifications from a specific forum category, he would receive a notification for every post and every reply to every post in that category. This is now fixed by differentiating between a post and a reply to a post.

Moreover, when a user makes a new post he can now specify whether he wants to receive notifications when someone replies to his post. 

For now, I use the user's first name. In a future task, this will be changed to be a unique username chosen by the user instead.


app/src/main/java/com/github/h3lp3rs/h3lp/forum/CachedForum.kt
```diff
@@ -33,9 +33,9 @@ class CachedForum(private val forum: Forum) : Forum {
     data class CacheHeader(val categoryPaths: ArrayList<String>)
 
     @RequiresApi(Build.VERSION_CODES.O)
-    override fun newPost(author: String, content: String): CompletableFuture<ForumPost> {
+    override fun newPost(author: String, content: String, isPost: Boolean): CompletableFuture<ForumPost> {
```
AlSchlo, line 36: I would just add a more relevant function name. 
Indeed newPost and isPost is a bit contradictory!
But don't change it, now we're used to it.

AlSchlo: Also, this file gets deleted by me (PR), to increase the coverage a bit. I'm keeping it under my sleeve for next sprint (integration)

app/src/androidTest/java/com/github/h3lp3rs/h3lp/forum/ForumNewPostActivityTest.kt
```diff
@@ -60,6 +60,8 @@ class ForumNewPostActivityTest {
 
         onView(withId(R.id.newPostSaveButton)).perform(ViewActions.scrollTo(),ViewActions.click())
 
+        onView(withId(R.id.switch_enable_notifications)).perform(ViewActions.scrollTo(),ViewActions.click())
```
Lindt19, line 63: You can import ViewActions

Lindt19 (approved):
Congratulations for fixing the bug ! Nice that notifications are working perfectly now 👍🏻

PR #270, opened 2022-05-16
250 forum mocking

AlSchlo:
This PR addresses #250 and mocks the forum using our mocked database.

Though this task may sound simple, it was an absolutely nightmare. Indeed, our mocked database and our firedatabase don't have exactly the same behaviour. This means that some modifications had to be made to make the database compatible.

Additionnaly, several bugs in the forum and in the mocked database were discovered.

Tests that mocked the forum using mockito were removed, and replaced with the new custom mocked forum. These all pass (the previous version did no real testing) and were also refactored on the way.

The mocked forum has been thoroughly tested. Only the listeners remain to be tested, **so beware when you use them**. This will be addressed during next sprint, this task already took way longer than expected, mainly due to all the debugging.

The forum was also refactored in new files to make it more coverage effective. 

All in all, our codebase will be a bit more robust if you accept this PR! :)


app/src/androidTest/java/com/github/h3lp3rs/h3lp/forum/ForumAnswersActivityTest.kt
```diff
@@ -1,83 +1,68 @@
...
+            selectedPost = post
 
-        onView(withId(R.id.add_answer_button)).perform(click())
+            launch<ForumAnswersActivity>(launchIntent).use {
+                // Add an answer to post
+                onView(withId(R.id.text_view_enter_answer)).perform(replaceText(ANSWER_TEST))
+                onView(withId(R.id.add_answer_button)).perform(click())
+
+                // No other way to do this, since delayed futures are not supported by Kotlin
+                // --> java.lang.NoSuchMethodError when using a delayed executor
```
Adrinels, line 55: That's some bad news :, (
one solution I see is that since this file contains only one test  we could code this test in java but that will not be really coherent

app/src/main/java/com/github/h3lp3rs/h3lp/forum/data/ForumPostData.kt
```diff
@@ -15,7 +15,7 @@ import java.time.ZonedDateTime
 data class ForumPostData(
     val author: String,
     val content: String,
-    val postTime: ZonedDateTime,
+    val postTime: String,
```
Adrinels, line 18: what is the  benefit of this change?

We lose some information with this. Someon posting message now in tawian will appeared to be send in the future for a person in the us 

The advantage of ZOnedDateTime is that it adapt the time with the zone the user is in

app/src/main/java/com/github/h3lp3rs/h3lp/forum/CachedForum.kt
```diff
@@ -1,179 +0,0 @@
...
-    }
-
-    override fun parent(): Forum {
-        return CachedForum(forum.parent())
-    }
-
-    companion object {
-        private const val CACHE_HEADER = "HEADER"
-    }
-}
```
Adrinels, line 179: Why did you deleted the cached forum ?

Adrinels (approved):
Globally a really good code .

I just write you some interrogation i have on certain poit

Well done!

app/src/main/java/com/github/h3lp3rs/h3lp/database/Databases.kt
```diff
@@ -15,7 +15,7 @@ import com.github.h3lp3rs.h3lp.storage.Storages.Companion.storageOf
  * Enumeration of all useful databases in H3LP
  */
 enum class Databases {
-    PREFERENCES, EMERGENCIES, NEW_EMERGENCIES, MESSAGES, CONVERSATION_IDS, PRO_USERS, FORUM;
+    PREFERENCES, EMERGENCIES, NEW_EMERGENCIES, MESSAGES, CONVERSATION_IDS, PRO_USERS;
```
Ahayderi, line 18: I guess there's some specific reason why you removed the forum database but I don't really understand it.

app/src/main/java/com/github/h3lp3rs/h3lp/database/MockDatabase.kt
```diff
@@ -94,13 +94,15 @@ class MockDatabase : Database {
...
-        concurrentLists.get(key)?.let { list ->
+        val list = concurrentLists[key]
+        if(list != null) {
             val gson = Gson()
             val convertedList = list.map { gson.fromJson(it, type)}
             future.complete(convertedList)
-        }?.run {
-            future.completeExceptionally(NoSuchFieldException("Key: $key not in the database"))
+        } else {
+            future.complete(emptyList())
```
Ahayderi, line 103: nice catch!

app/src/main/java/com/github/h3lp3rs/h3lp/forum/ForumCategory.kt
```diff
@@ -23,17 +29,31 @@ enum class ForumCategory {
         fun forumOf(choice: ForumCategory): Forum {
             // The categories are the first sub-level of the forum, thus they lie after root (the
             // empty list)
-            choice.forum = choice.forum ?: FireForum(listOf(choice.name))
-            return choice.forum!!
+            root?.let {
+                return it.child(choice.name)
+            }
```
Ahayderi, line 34: same remark as above, I'm not sure why you use directly FireDatabase(ROOT_FORUM_DB_PATH) instead of databaseOf(FORUM) as we had beforehand since this seems like an abstraction leak. We should at least have the choice to choose the db to use in the forum if we switched from Firebase (for example after having had too many bad experiences with it :))

app/src/main/java/com/github/h3lp3rs/h3lp/forum/ForumCategory.kt
```diff
@@ -23,17 +29,31 @@ enum class ForumCategory {
         fun forumOf(choice: ForumCategory): Forum {
             // The categories are the first sub-level of the forum, thus they lie after root (the
             // empty list)
-            choice.forum = choice.forum ?: FireForum(listOf(choice.name))
-            return choice.forum!!
+            root?.let {
+                return it.child(choice.name)
+            }
+            root = FireDBForum(emptyList(), FireDatabase(ROOT_FORUM_DB_PATH))
+            return root!!.child(choice.name)
```
Ahayderi, line 36: an alternative to avoid the !! operator since it's usually bad practice:
```suggestion
            val forum = FireDBForum(emptyList(), FireDatabase(ROOT_FORUM_DB_PATH))
            root = forum
            return forum.child(choice.name)
```

app/src/main/java/com/github/h3lp3rs/h3lp/forum/ForumCategory.kt
```diff
@@ -23,17 +29,31 @@ enum class ForumCategory {
         fun forumOf(choice: ForumCategory): Forum {
```
Ahayderi, line 29: forumOf could be changed into a simple call to root().child(choice.name) (this would avoid the code duplication too)

app/src/androidTest/java/com/github/h3lp3rs/h3lp/forum/ForumAnswersActivityTest.kt
AlSchlo: Probably there is a way to update Kotlin guava or something like that to make these classes work. But that's a bit risky to do one sprint before the final deadline :(

app/src/main/java/com/github/h3lp3rs/h3lp/forum/CachedForum.kt
AlSchlo: Coverage since not yet integrated.

app/src/main/java/com/github/h3lp3rs/h3lp/forum/ForumCategory.kt
```diff
@@ -23,17 +29,31 @@ enum class ForumCategory {
...
+         * Returns the root of the forum
          */
-        fun setForum(choice: ForumCategory, newForum: Forum) {
-            choice.forum = newForum
+        fun root(): Forum {
+            root?.let {
+                return it
+            }
+            root = FireDBForum(emptyList(), FireDatabase(ROOT_FORUM_DB_PATH))
+            return root!!
```
Ahayderi, line 47: same idea as above to avoid the "bad practice" of !!
```suggestion
            rootForum = FireDBForum(emptyList(), FireDatabase(ROOT_FORUM_DB_PATH))
            root = rootForum
            return rootForum
```

app/src/main/java/com/github/h3lp3rs/h3lp/forum/ForumPost.kt
```diff
@@ -40,7 +40,7 @@ class ForumPost(
      * @return this This post updated with potential replies from others
      * in the form of a future
      */
-    private fun refresh() : CompletableFuture<ForumPost> {
+    fun refresh() : CompletableFuture<ForumPost> {
```
Ahayderi, line 43: Good idea, this could be very useful!

app/src/main/java/com/github/h3lp3rs/h3lp/forum/implementation/SimpleDBForum.kt
```diff
@@ -1,267 +1,268 @@
...
+    }
+
+    /**
+     * Gets the post date in the following format day of month month hour-minutes
+     * @param currentTime The current date
+     * @return Formatted current date
+     */
+    @RequiresApi(Build.VERSION_CODES.O)
+    private fun getFormattedPostTime(currentTime: ZonedDateTime): String {
+        return currentTime.dayOfMonth.toString() + "" + currentTime.month.toString() + " " +
```
Ahayderi, line 69: \+ "" seems bizarre :monocle_face:

Ahayderi (approved):
Very nice implementation of a mock forum, I can only imagine the trouble you went through to make this work so I'm very impressed by what you achieved! There are some changes I didn't fully understand so I'm eager to discuss further, keep up the good work!

app/src/main/java/com/github/h3lp3rs/h3lp/database/Databases.kt
AlSchlo: It shouldn't be accessible via the Databases enumeration!
If we call databaseOf(FORUM) somewhere in the code, we risk breaking stuff. The forum interface should be the only point of access.

app/src/main/java/com/github/h3lp3rs/h3lp/forum/ForumCategory.kt
AlSchlo: My philosophy was the following:
- Databases are for, well, databases
- The forum is implemented with a database (here Firebase). It is not a database per sea, just the support. 
- Cf. previous remark
- If we want to switch from Firebase, we should be able to keep the SimpleDBForum abstract class that is Database agnostic, so I don't see the issue here.
Though I get your point, I still believe that having the public available support of the forum (databaseOf(FORUM)) accessible a bad idea.

app/src/main/java/com/github/h3lp3rs/h3lp/forum/implementation/SimpleDBForum.kt
AlSchlo: Not my code, cherry picked from Wiam's commit :D

app/src/main/java/com/github/h3lp3rs/h3lp/forum/ForumCategory.kt
AlSchlo: I want to push this, but I won't waste a whole Cirrus build for this.

app/src/main/java/com/github/h3lp3rs/h3lp/forum/ForumCategory.kt
AlSchlo: I want to push this, but I won't waste a whole Cirrus build for this.

app/src/main/java/com/github/h3lp3rs/h3lp/forum/implementation/SimpleDBForum.kt
wiam8: @Ahayderi don't worry, I noticed it. It has been fixed on my PR

PR #272, opened 2022-05-19
236 design refont

Adrinels:
This PR refont the app design, fix some UI bug and adapt the theme to the new logo color.

SInce this PR is mainly xml file I recomend you to open the app on your avd and check the design directly instead of reviewing the xml file which can be kind of a pitty.

I am open to every remark and if you prefer some design element of the old version I would be happy to change it.


emduc (approved):
Breathtaking ! Well done !
Since those are pure UI changes I find it easier to summarize everything here:

- the help parameters are absolutely beautiful but do not work anymore :( this absolutely needs to be fixed
- I think you missed the latest message activity layout
- I'm not a huge fan of the colours for the button in the emergency call pop-up, red + green is a little bit aggressive imo, what about grey for cancel and red for call?
- I know you have not changed it yourself, but the title of the emergency_call_option_pop_up seems a bit off, it's the same string as in the medical card information and I am not sure it's on purpose

Apart from those few details, it's once again very impressive, nice work Adrien

Ahayderi (approved):
Your PRs never disappoint! Your attention to detail is a real pleasure, well done. I've put a few notes below since this is UI, but these are minor details and overall, this new look on the app is amazing!

- I really like the font but there seems to be a problem for certain characters which are not parsed well as you can see on the below image (now, this might be a problem from my end because my computer might not have this font installed so let me know if that's the case)
image
- As noted by Emilien, the main problem is the buttons in the HelpeeSelectionActivity not working anymore
- This was already mentioned in the scrum meeting but it would be great to add a task to change the rating activity to fit with the current theme
- On the AwaitHelpActivity, the text for the messages and for the emergency call overlap for some reason as you can see below
image

PR #274, opened 2022-05-19
254 code review

Ahayderi:
This PR tackles #254. There's a **lot** of modified files but most of the changes are trivial. I ended up implementing more than the 60-70% I projected but there's a few time consuming tasks remaining.
Note: I know the tests don't pass yet but I know what causes the problem and will thus modify it as soon as I'm back :D
The list of modifications is below, everything marked with an X at the beginning was implemented:

### `AwaitHelpActivity`
- X `setupLocation` documentation refers to an invalid method.
- X See comment on `FirstAidActivity` for the functions at the end of the class.
- X Missing comments on some functions (`foundHelperPerson`, `showHelperPerson`)
### `FirstAidActivity`
- X We propose to refactor the activity in the following way: 
    - Create a map, mapping the button id to an intent going to the Activity it refers to.
    - Having a single function goToActivity(v:View) that uses `v.id` to launch the correct activity. 
    - this way, we avoid the code duplication you have at the moment.
### `GoogleAPIHelper`
- X `downloadURL` missing @return comment.
- X (`displayWalkingPath`, line 51) We detect some inconsistency here. Refers to the comment on `JSONParserInterface`. We encourage you to return `null` value from the JSON parser if an error happen. To handle the `null` value, instead of `path.let` you can use `path?.let` and in the function you can use `it` instead of `path`.
- X `parseTask` missing @return and @param comments and return type should be changed to `T?`.
### `LocalEmergencyCaller`
- X `getPhoneNumberFromCountry` the reader is not closed properly `csvReader.close()`.
### `MainPageActivity`
- X See comment on `FirstAidActivity` for the functions at the end of the class. 
- Missing comments on the `MainPageButton`.
### MapsFragment
- X `showMap`: since the hashmap place is a `HashMap<String, String>`, the returned value cannot be `null` and therefore the two last conditions on the first if are not needed.
### `MedicalCardActivity`
- X `createBirthField`, `createHeightField`, `createWeightField`, `createBloodField`, `createGenderField`, `createHelpField`: the content of these methods could be directly replaced in the onCreate() function (keep their comments though) since they are called once and contains only one line on code. 
- X The two `loadTo` methods don't have comments.
- X Missing @return comment for `checkField` and `checkPolicy` functions.
- X Missing comments for `getStringFromId` and `getIntFromId`

### `MySkillsActivity`
- X `getBooleanFromSwitch`, `toggleSwitch` missing comments.
- X  Remove the annotation after having upgraded the SDK version

### `NearbyUtilitiesActivity`
- X Missing comments `setRequestedButton` and `setupLocation`
### `EmergencyInfoRepository`
- X @param for `database` is missing.
### `Repositories`
- It might be better to wrap everything in an `object` instea of having "global variable".
### `Database`
- The signature of the `addListener method should change (according to the comments we will make in the `FireDatabase` class).
### `Databases`
- X @return missing for the return type of databaseOf method.
### `FireDatabase`
- X In the JavaDoc comment of the get method, should have Long instead of Int. Missing comment for the returned value.
- X The comments of the overidden functions should not be rewritten in sub-classes/interface implementation.
- X Missing description for the `path` parameter of the class.
- To be consistent with the `get` function, it would be nice to also the `reified` keyword instead of explicitly asking the type as a parameter of the function, since the class.java can be obtained from the `T` type (using `T::class.java`).
- X In the same function, for the `onCancelled` case, you should write the error in the log (using `Log.e(...)`) instead of printing a message in the console.

### `MockDatabase`
- This class should be in the test package as it is not used during the normal runtime of the application.
- The method `addListener` will change according to the comments made above.
### `HelperSkills`
- X No comments for the different values of the data class.
### `MedicalInformation`
- X No comment for the different values of the data class.
- X You should upgrade the API version of your app instead of having `RequiresAPI(...)` annotation.
- X The `MIN_WEIGHT` cannot be set to 20 if the birthdate can be today. However, for legal purposes (ask us in the next meeting for more information), it should be preferred to limit the age of the app user. 
- X Small typo in the `BloodType` enum, `Unkown` should be replaced by `Unknown`.
## Package `h3lp.firstaid`
- Instead of having 4 different activities with the same structure/same code and different content, you could have only one activity which takes a parameter [(more info)](https://stackoverflow.com/questions/2405120/how-to-start-an-intent-by-passing-some-parameters-to-it) and adapt the content depending on the parameter. This avoid code duplication and is better practice. 
### `GeneralLocationManager`
- CHOSE NOT TO IMPLEMENT you did the get/set property is fine. However, for more concise notation, see [properties documentation](https://kotlinlang.org/docs/properties.html#getters-and-setters) of Kotlin.
- X Minor issue but it could be nice to change the name of the last function from `setSystemManager` to `setDefaultSystemManager`. 
### `LocationManagerInterface`
- X Missing comment for return value
### `SystemLocationAdapter`
- X No need to comment the function again since it is an overidden function and the comment is the same.
### `NotificationService`
- X `createNotificationChannel`, you can upgrade the SDK version of Android to avoid this if.
- In `sendOpenActivityNotification`, you can use `reified` keyword instead of explicitly giving the type as parameter.
- X `buildBasicNotification`, minor typos in the @return of the function.
### `PresIrrelevantActivity`
- X One improvement could be to change the color of the button when we agree to the terms of use.
- X Missing comment for `addClickableText`
### `SwipeListener`
- X Missing comments in the companion object.
### `GoogleSignInAdapter`
- X No need to comment again for overidden functions.
### `SignIn`
- X You can use properties as suggested before. If you don't want to, you should comment the set function.
### `SignInActivity`
- :eyes: did not have the time nor courage to implement this As said in the general remark, storing context is a bad practice [(see this)](https://stackoverflow.com/questions/37709918/warning-do-not-place-android-context-classes-in-static-fields-this-is-a-memory). Therefore, the context should be removed from the companion object.
- X Missing comment for the functions `checkToSAndLaunchIfNotAcceptedElseMain`
- X If `signInClient` is not used elsewhere, should be private, otherwise you should comment what it represents.  
### `SignInInterface`
- X Really nice abstraction
- X Missing @return comment for `isSignedIn`.
### `LocalStorage`
- You should throw a more explicit Error (e.g. `UserNotAuthenticatedException`) that will be catched later on. 
- Same remark as earlier for the `context` object stored in the instance of the class.
- X Missing @param comment for `parseOnlinePrefs`.
- Extract one function containing the db object creation, that will throw a more meaningful error if the user is not logged in.
- X The comments (@param, @return) are missing in this class for almost all the methods :). 
- **Still need to put futures for get methods** These methods need to be implemented asynchronously since disk I/O are slow operation that should not be blocking and to be consistent with what you did for the online database.
- `getObjectOrDefault` and `setObject` should use `reified` keyword instead of explicit class type.
### `Storages`
- Same remark as earlier for the `context` object stored in the instance of the class.
- X `storageOf` missing @return comment.
- X Missing comment for class parameter.
### `AedLocations`
- The value `AED_LOCATIONS_LAUSANNE` should be wrapped in an object. Also, the aed locations should be store in the online database and retrieve when the application needs it and eventually cached locally (since the location is unlikely to be changed too often).
### `JSONParserInterface`
- X `parseResult` should return `T?` instead of `T` otherwise you have no way of knowing if there was an error during the parsing.
The other classes should be adapted with respect to the above remark. They should also include comments :)


app/src/main/java/com/github/h3lp3rs/h3lp/database/FireDatabase.kt
```diff
@@ -9,13 +10,15 @@ import java.util.concurrent.CompletableFuture
...
  */
-internal class FireDatabase(path : String) : Database {
+internal class FireDatabase(path: String) : Database {
 
-    private val db : DatabaseReference =
-        Firebase.database("https://h3lp-signin-default-rtdb.europe-west1.firebasedatabase.app/").reference.child(
-            path
-        )
+    private val db: DatabaseReference = Firebase
+        .database("https://h3lp-signin-default-rtdb.europe-west1.firebasedatabase.app/")
```
Adrinels, line 20: Just wondering if we shouldn't put this url in a value files in res folder

app/src/main/java/com/github/h3lp3rs/h3lp/dataclasses/HelperSkills.kt
```diff
@@ -1,4 +1,12 @@
 package com.github.h3lp3rs.h3lp.dataclasses
 
+/**
```
Adrinels, line 3: Could add some quick description of the use of the class

app/src/main/java/com/github/h3lp3rs/h3lp/dataclasses/MedicalInformation.kt
```diff
@@ -7,7 +7,17 @@ import com.google.i18n.phonenumbers.Phonenumber
 import java.lang.IllegalArgumentException
 import java.time.Year
 
-@RequiresApi(Build.VERSION_CODES.O)
+/**
```
Adrinels, line 10: Same here

app/src/main/java/com/github/h3lp3rs/h3lp/signin/GoogleSignInAdapter.kt
```diff
@@ -19,35 +19,26 @@ import java.util.*
...
  * It is used as the central sign in interface in the app
  */
-object GoogleSignInAdapter: SignInInterface<AuthResult> {
+object GoogleSignInAdapter : SignInInterface<AuthResult> {
     var auth: FirebaseAuth = Firebase.auth
     lateinit var gso: GoogleSignInOptions
 
     private const val SERVER_CLIENT_ID =
-    "899579782202-t3orsbp6aov3i91c99r72kc854og8jad.apps.googleusercontent.com"
+        "899579782202-t3orsbp6aov3i91c99r72kc854og8jad.apps.googleusercontent.com"
```
Adrinels, line 27: we could then put this url int the url value file too.

Then somebody wanting to use our project with other credential can easily replace our url by his

Adrinels (commented):
Nice work!

Nothing special to say otherwise that it is a good refactor. 

Just some interogation how we store url in our project

app/src/main/java/com/github/h3lp3rs/h3lp/AwaitHelpActivity.kt
```diff
@@ -248,6 +247,12 @@ class AwaitHelpActivity : AppCompatActivity() {
         startActivity(intent)
     }
 
+    /**
+     * Launches the RecentMessagesActivity with the given emergency key to display the messaging
+     * screen
```
AlSchlo, line 252:
```suggestion
     * screen.
```

app/src/main/java/com/github/h3lp3rs/h3lp/FirstAidActivity.kt
```diff
@@ -21,23 +33,13 @@ class FirstAidActivity : AppCompatActivity() {
...
-    fun goToAsthmaActivity(view: View) {
-        goToActivity(AsthmaActivity::class.java)
+    /**
+     * Called whenever a user clicks a button in the activity, launches the button's corresponding
+     * activity (as defined in buttonToActivity)
+     * @param view The button that was clicked
+     */
+    fun goToButtonActivity(view: View) {
+        // If the view isn't one of the buttons, don't do anything
+        buttonToActivity[view]?.let { goToActivity(it) }
```
AlSchlo, line 43: Using a `contains` would look better and be more readable.

app/src/main/java/com/github/h3lp3rs/h3lp/MainPageActivity.kt
```diff
@@ -481,14 +473,18 @@ class MainPageActivity : AppCompatActivity(), OnRequestPermissionsResultCallback
...
\ No newline at end of file
+/**
+ * Data class that represents the main page buttons and some information about them (the fact that
+ * they are in the scroll view as well as the prompt that should appear for them in the guide) to be
+ * able to directly get the information on that button
+ * @param buttonId The button's id
+ * @param isInScrollView If the button appears in the scroll view (the scrolling window in the middle
+ * of the main page activity)
+ * @param promptTextId The text that should be shown when the guide does an explanation for that
+ *  button
```
AlSchlo, line 484:
```suggestion
 * button
```

app/src/main/java/com/github/h3lp3rs/h3lp/MedicalCardActivity.kt
```diff
@@ -71,7 +96,7 @@ class MedicalCardActivity : AppCompatActivity() {
 
 
     /**
-     * initialises the field for emergency contact phone number so that we make
+     * Initialises the field for emergency contact phone number so that we make
```
AlSchlo, line 99:
```suggestion
     * Initializes the field for emergency contact phone number so that we make
```

app/src/main/java/com/github/h3lp3rs/h3lp/MedicalCardActivity.kt
```diff
@@ -106,13 +131,13 @@ class MedicalCardActivity : AppCompatActivity() {
...
 
     /**
-     * create an Field that test input and write error 7
-     * @param idEditText the id of the editText to test value
-     * @param idTextInputLayout the Layout where to display erro
-     * @param min the min margin
-     * @param max the max margin
-     * @param minErrorMsg the message to display if smallest than min
-     * @param maxErrorMsg the message to display if biggest than min
+     * Create an Field that test input and write error 7
```
AlSchlo, line 134:
```suggestion
     * Creates a field that tests the input and writes error 7
```

app/src/main/java/com/github/h3lp3rs/h3lp/MedicalCardActivity.kt
```diff
@@ -106,13 +131,13 @@ class MedicalCardActivity : AppCompatActivity() {
...
+     * @param idEditText The id of the editText to test value
+     * @param idTextInputLayout The Layout where to display erro
```
AlSchlo, line 136:
```suggestion
     * @param idTextInputLayout The layout where to display the error
```

app/src/main/java/com/github/h3lp3rs/h3lp/database/FireDatabase.kt
```diff
@@ -24,11 +27,13 @@ internal class FireDatabase(path : String) : Database {
      * Only works on the following types (due to Firebase's policy):
      * - Boolean
      * - String
-     * - Int
+     * - Long
```
AlSchlo, line 30: Except that this is a lie.
Of course Firebase works with `Longs`, but here we adapted it such that it can only handle Int/String/Boolean (to satisfy **our** database requirements). Indeed, `Ints` are more used than `Longs` usually.

app/src/main/java/com/github/h3lp3rs/h3lp/database/FireDatabase.kt
```diff
@@ -39,43 +44,44 @@ internal class FireDatabase(path : String) : Database {
...
     }
 
-    override fun setString(key : String, value : String) {
+    override fun setString(key: String, value: String) {
         db.child(key).setValue(value)
     }
 
-    override fun getDouble(key : String) : CompletableFuture<Double> {
+    override fun getDouble(key: String): CompletableFuture<Double> {
         // This Fix is due to a misconception in firebase:
```
AlSchlo, line 64: Consider /* */

app/src/main/java/com/github/h3lp3rs/h3lp/database/repositories/EmergencyInfoRepository.kt
```diff
@@ -6,6 +6,7 @@ import java.util.concurrent.CompletableFuture
 
 /**
  * Repository for emergency information communication with the database
+ * @param database The database from which to create the repository
  */
 class EmergencyInfoRepository(private val database: Database) : Repository<EmergencyInformation>{
```
AlSchlo, line 11: By the way, this class is not really useful. It could be removed when I think about it. It doesn't offer a lot of useful database wrapping.

app/src/main/java/com/github/h3lp3rs/h3lp/database/repositories/Repository.kt
```diff
@@ -3,27 +3,26 @@ package com.github.h3lp3rs.h3lp.database.repositories
 import java.util.concurrent.CompletableFuture
 
 /**
- * General interface displaying the different methods required to store, retrieve and delete an object of T in the database
+ * General interface displaying the different methods required to store, retrieve and delete an
+ * object of T in the database
  */
-interface Repository<T>{
+interface Repository<T> {
```
AlSchlo, line 9: Same as above. Would increase coverage as well I believe. This class does not score **A** on code climate and is  kind of useless.

AlSchlo (approved):
Your refactoring is like a breath of fresh air. I like your coding style.

The only thing I disapprove is the documentation of private functions. This makes the file longer and takes more effort for us, developers, and is rarely of great interest.

app/src/main/java/com/github/h3lp3rs/h3lp/AwaitHelpActivity.kt
AlSchlo: Definitely not nitpicking 🐦

Adrinels (approved):

app/src/main/java/com/github/h3lp3rs/h3lp/database/FireDatabase.kt
Ahayderi: Done

app/src/main/java/com/github/h3lp3rs/h3lp/AwaitHelpActivity.kt
Ahayderi: Obviously not :) (also the dot doesn't follow our comments convention.. so I guess this is counter-nitpicking).

app/src/main/java/com/github/h3lp3rs/h3lp/database/repositories/EmergencyInfoRepository.kt
Ahayderi: This can be discussed at our next meeting but good point indeed

PR #284, opened 2022-05-20
fix the buttons

Adrinels:
This PR fix two problem on button of the UI:

1. the buttons in helpee parameters did not work anymore.

2. the chat button in await help activity not appearing when no helper is comming is not really nice in the layout.

Good review


emduc (approved):
Thanks for that !

app/src/main/java/com/github/h3lp3rs/h3lp/AwaitHelpActivity.kt
```diff
@@ -74,7 +75,8 @@ class AwaitHelpActivity : AppCompatActivity() {
 
             // Initially the contact helpers is hidden, only after a user responds to the request it
             // becomes visible.
-            constraint_layout_contact_helpers.visibility = View.INVISIBLE
+            image_open_latest_messages.setColorFilter(ContextCompat.getColor(this, R.color.light_gray), android.graphics.PorterDuff.Mode.SRC_IN)
```
Lindt19, line 78: Maybe the color can be put in a variable since it's used multiple times.

Lindt19 (approved):
Thank you for the fix 👍🏻 Glad it's working perfectly now

PR #302, opened 2022-05-25
275 cache integration

AlSchlo:
**_WARNING_**: This PR reuses branch #276 . So if you haven't checked that one yet, do it first (and wait for a merge into main) to reduce the amount of files to review. Indeed, I had to use the `MockDBForum` implementation to test the cache efficiently.

This PR tackles #275 and finally integrates a cache into our app!

The design is very elegant: I simply coded a forum wrapper cache that allows us to cache any `SimpleDBForum` instance **on the fly**. As a consequence, the integration itself was extremely easy.

Of course, as usual, this PR contains some bug fixes to the `SimpleDBForum` implementation, particularly regarding the `getAll` function.

Testing the cache went smoothly with the brand new `MockDBForum` and as a result, the coverage once again increased by a solid **0.7%**! 

The cache truely is a pineapple 🍍  of technology, now guests can look up their most pressing issues **offline**! As a consequence, I **strongly** recommend you to accept this PR.

_It has been a tremendous pleasure working with you.
Thank you for the wonderful team!_

Yours faithfully, 

AlSchlo

image

### DIFF
--- a/app/src/androidTest/java/com/github/h3lp3rs/h3lp/forum/ForumCacheTest.kt
+++ b/app/src/androidTest/java/com/github/h3lp3rs/h3lp/forum/ForumCacheTest.kt
@@ -109,10 +109,13 @@ class CacheForumTest {
         postOnline.reply(AUTHOR, CONTENT).join()
 
         // Fetch through cache the post
-        val all = cachedForum.getAll().join()
+        cachedForum.getAll().join()
 
         // Reset the forum, acts as a clear of the data
         resetOnlineForum()
+
+        // Fetch through cache the post
+        val all = cachedForum.getAll().join()
 
         assertEquals(1, all.size)
         assertEquals(1, all[0].second.size)
@@ -129,11 +132,13 @@ class CacheForumTest {
             post.reply(AUTHOR, CONTENT).join()
         }
 
-        // Fetch through cache the post
-        val all = cachedForum.root().getAll().join()
+        cachedForum.root().getAll().join()
 
         // Reset the forum, acts as a clear of the data
         resetOnlineForum()
+
+        // Fetch through cache the post
+        val all = cachedForum.root().getAll().join()
 
         assertEquals(ForumCategory.values().size, all.size)
         for(c in all) {
@@ -144,8 +149,8 @@ class CacheForumTest {
 
     @Test
     fun listenerAddsToCache() {
-        // Do nothing in basic listener
-        cachedForum.root().listenToAll {  }
+        var counter = 0
+        cachedForum.root().listenToAll { counter++ }
 
         // Add an external post
         val post = rawForum.newPost(AUTHOR, CONTENT, isPost = true).join()
@@ -157,6 +162,27 @@ class CacheForumTest {
 
         assertEquals(AUTHOR, cachedPost.post.author)
         assertEquals(CONTENT, cachedPost.post.content)
+        assertEquals(1, counter)
+    }
+
+    @Test
+    fun replyListenerAddsToCache() {
+        var counter = 0
+        cachedForum.root().listenToAll { counter++ }
+
+        // Add an external post
+        val post = rawForum.newPost(AUTHOR, CONTENT, isPost = true).join()
+        post.reply(AUTHOR, CONTENT).join()
+
+        // Reset the forum, acts as a clear of the data
+        resetOnlineForum()
+
+        val all = cachedForum.getAll().join()
+
+        assertEquals(1, all.size)
+        assertEquals(1, all[0].second.size)
+        assertEquals(1, all[0].second[0].replies.size)
+        assertEquals(2, counter)
     }
 
     companion object {
--- a/app/src/androidTest/java/com/github/h3lp3rs/h3lp/forum/ForumCacheTest.kt
+++ b/app/src/androidTest/java/com/github/h3lp3rs/h3lp/forum/ForumCacheTest.kt
@@ -1,0 +1,149 @@
+package com.github.h3lp3rs.h3lp.forum
+
+import androidx.test.core.app.ApplicationProvider
+import com.github.h3lp3rs.h3lp.database.Databases.*
+import com.github.h3lp3rs.h3lp.database.Databases.Companion.setDatabase
+import com.github.h3lp3rs.h3lp.database.MockDatabase
+import com.github.h3lp3rs.h3lp.forum.implementation.CachedForum
+import com.github.h3lp3rs.h3lp.signin.SignInActivity.Companion.globalContext
+import com.github.h3lp3rs.h3lp.storage.Storages.Companion.resetStorage
+import org.junit.Assert.assertEquals
+import org.junit.Before
+import org.junit.Test
+
+class CacheForumTest {
+
+    // Useful variables
+    private lateinit var rawForum: Forum
+    private lateinit var cachedForum: Forum
+
+    @Before
+    fun setup() {
+        globalContext = ApplicationProvider.getApplicationContext()
+        resetStorage()
+        setDatabase(PREFERENCES, MockDatabase())
+        ForumCategory.mockForum()
+        rawForum = ForumCategory.root().child(ForumCategory.GENERAL.name)
+        cachedForum = CachedForum(rawForum)
+    }
+
+    private fun resetOnlineForum() {
+        ForumCategory.mockForum()
+        rawForum = ForumCategory.root().child(ForumCategory.GENERAL.name)
+        cachedForum = CachedForum(rawForum)
+    }
+
+    @Test
+    fun newPostIsAddedToCache() {
+        // Add a post
+        val postOnline = cachedForum.newPost(AUTHOR, CONTENT, isPost = true).join()
+
+        // Reset the forum, acts as a clear of the data
+        resetOnlineForum()
+
+        // Fetch through cache the post
+        val postCached = cachedForum.getPost(postOnline.post.key).join()
+
+        assertEquals(CONTENT, postCached.post.content)
+        assertEquals(AUTHOR, postCached.post.author)
+    }
+
+    @Test
+    fun repliesAreAddedToCache() {
+        // Add a post
+        val postOnline = cachedForum.newPost(AUTHOR, CONTENT, isPost = true).join()
+
+        // Add two replies
+        postOnline.reply(AUTHOR, CONTENT).join()
+        postOnline.reply(AUTHOR, CONTENT).join()
+
+        // Reset the forum, acts as a clear of the data
+        resetOnlineForum()
+
+        // Fetch through cache the post
+        val postCached = cachedForum.getPost(postOnline.post.key).join()
+
+        assertEquals(2, postCached.replies.size)
+    }
+
+    @Test
+    fun postAndGetDoesNotDuplicate() {
+        // Add a post
+        val postOnline = cachedForum.newPost(AUTHOR, CONTENT, isPost = true).join()
+
+        // Fetch it again
+        cachedForum.getPost(postOnline.post.key).join()
+
+        // Reset the forum, acts as a clear of the data
+        resetOnlineForum()
+
+        // Fetch through cache the post
+        val all = cachedForum.getAll().join()
+
+        assertEquals(1, all.size)
+        assertEquals(1, all[0].second.size)
+    }
+
+    @Test
+    fun getAddsToCache() {
+        // Add an external post
+        val postOnline = rawForum.newPost(AUTHOR, CONTENT, isPost = true).join()
+
+        // Fetch it through the cache
+        cachedForum.getPost(postOnline.post.key).join()
+
+        // Reset the forum, acts as a clear of the data
+        resetOnlineForum()
+
+        // Fetch through cache the post
+        val cachedPost = cachedForum.getPost(postOnline.post.key).join()
+
+        assertEquals(CONTENT, cachedPost.post.content)
+        assertEquals(AUTHOR, cachedPost.post.author)
+    }
+
+    @Test
+    fun getAllAddsToCache() {
+        // Add an external post and reply
+        val postOnline = rawForum.newPost(AUTHOR, CONTENT, isPost = true).join()
+        postOnline.reply(AUTHOR, CONTENT).join()
+
+        // Fetch through cache the post
+        val all = cachedForum.getAll().join()
+
+        // Reset the forum, acts as a clear of the data
+        resetOnlineForum()
+
+        assertEquals(1, all.size)
+        assertEquals(1, all[0].second.size)
+        assertEquals(1, all[0].second[0].replies.size)
+    }
+
+    @Test
+    fun getAllFromDifferentCategoriesAddsToCache() {
+        // Add external posts and two replies
+        val rawRoot = rawForum.root()
+        for (category in ForumCategory.values()) {
+            val post = rawRoot.child(category.name).newPost(AUTHOR, CONTENT, isPost = true).join()
+            post.reply(AUTHOR, CONTENT).join()
+            post.reply(AUTHOR, CONTENT).join()
+        }
+
+        // Fetch through cache the post
+        val all = cachedForum.root().getAll().join()
+
+        // Reset the forum, acts as a clear of the data
+        resetOnlineForum()
+
+        assertEquals(ForumCategory.values().size, all.size)
+        for(c in all) {
+            assertEquals(1, c.second.size)
+            assertEquals(2, c.second[0].replies.size)
+        }
+    }
+
+    companion object {
+        private const val AUTHOR = "AUTHOR"
+        private const val CONTENT = "CONTENT"
+    }
+}
--- a/app/src/androidTest/java/com/github/h3lp3rs/h3lp/forum/ForumCacheTest.kt
+++ b/app/src/androidTest/java/com/github/h3lp3rs/h3lp/forum/ForumCacheTest.kt
@@ -142,6 +142,23 @@ class CacheForumTest {
         }
     }
 
+    @Test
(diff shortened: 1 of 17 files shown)


app/src/main/java/com/github/h3lp3rs/h3lp/forum/implementation/CachedForum.kt
```diff
@@ -0,0 +1,172 @@
...
+    private fun setCacheEntry(relativePath: Path, entry: CacheEntry) {
+        val cachePath = (path + relativePath).joinToString(separator = "/")
+        cache.setObject(cachePath, CacheEntry::class.java, entry)
+    }
+
+    override fun listenToAll(action: (ForumPostData) -> Unit) {
+        val newWrappedAction: (ForumPostData) -> Unit = {
+            // This is a very hacky but easy way to know from the data, the path to its forum pointer
+            // We should change the listener design to make it cleaner, but for now it works fine
+            val path = listOf(it.category.name) + if(it.isPost) emptyList() else listOf(it.key)
```
Ahayderi, line 153: I'm pretty sure the path should be the following (for a reply, its path is on category/repliesKey, not category/forumPostKey
```suggestion
            val path = listOf(it.category.name) + if(it.isPost) emptyList() else listOf(it.repliesKey)

```

app/src/androidTest/java/com/github/h3lp3rs/h3lp/forum/ForumCacheTest.kt
```diff
@@ -0,0 +1,166 @@
...
+    private fun resetOnlineForum() {
+        ForumCategory.mockForum()
+        rawForum = ForumCategory.root().child(ForumCategory.GENERAL.name)
+        cachedForum = CachedForum(rawForum)
+    }
+
+    @Test
+    fun newPostIsAddedToCache() {
+        // Add a post
+        val postOnline = cachedForum.newPost(AUTHOR, CONTENT, isPost = true).join()
```
Ahayderi, line 39: as usual, Im not a fan of the .join() calls, instead of modifying all the tests, maybe just add a comment at the beginning of the tests stating that since we're using the mockDatabase which runs locally, the joins won't cause any problem

app/src/androidTest/java/com/github/h3lp3rs/h3lp/forum/ForumCacheTest.kt
```diff
@@ -0,0 +1,166 @@
...
+
+        // Add two replies
+        postOnline.reply(AUTHOR, CONTENT).join()
+        postOnline.reply(AUTHOR, CONTENT).join()
+
+        // Reset the forum, acts as a clear of the data
+        resetOnlineForum()
+
+        // Fetch through cache the post
```
Ahayderi, line 63:
```suggestion
        // Fetch the post through the cache

```

app/src/androidTest/java/com/github/h3lp3rs/h3lp/forum/ForumCacheTest.kt
```diff
@@ -0,0 +1,166 @@
...
+        rawForum = ForumCategory.root().child(ForumCategory.GENERAL.name)
+        cachedForum = CachedForum(rawForum)
+    }
+
+    private fun resetOnlineForum() {
+        ForumCategory.mockForum()
+        rawForum = ForumCategory.root().child(ForumCategory.GENERAL.name)
+        cachedForum = CachedForum(rawForum)
+    }
+
```
Ahayderi, line 35: As you can see in the comments in the CachedForum, if you tested a user posting on the cached forum, then getting the post they posted, they wouldn't actually see it (for the reasons mentioned below).

app/src/androidTest/java/com/github/h3lp3rs/h3lp/forum/ForumCacheTest.kt
```diff
@@ -0,0 +1,166 @@
...
+    fun setup() {
+        globalContext = ApplicationProvider.getApplicationContext()
+        resetStorage()
+        setDatabase(PREFERENCES, MockDatabase())
+        ForumCategory.mockForum()
+        rawForum = ForumCategory.root().child(ForumCategory.GENERAL.name)
+        cachedForum = CachedForum(rawForum)
+    }
+
+    private fun resetOnlineForum() {
```
Ahayderi, line 30: I don't think resetting the online forum is the "good test" for the cached forum, for me, you should switch to a forum that only returns failed futures (which is what would happen if the user lost their internet connection)

app/src/androidTest/java/com/github/h3lp3rs/h3lp/forum/ForumCacheTest.kt
```diff
@@ -0,0 +1,166 @@
...
+    fun getAllAddsToCache() {
+        // Add an external post and reply
+        val postOnline = rawForum.newPost(AUTHOR, CONTENT, isPost = true).join()
+        postOnline.reply(AUTHOR, CONTENT).join()
+
+        // Fetch through cache the post
+        val all = cachedForum.getAll().join()
+
+        // Reset the forum, acts as a clear of the data
+        resetOnlineForum()
```
Ahayderi, line 115: shouldn't you call getAll again after resetting the online forum (you don't seem to be testing anything here since all takes value before the reset)

app/src/androidTest/java/com/github/h3lp3rs/h3lp/forum/ForumCacheTest.kt
```diff
@@ -0,0 +1,166 @@
...
+            val post = rawRoot.child(category.name).newPost(AUTHOR, CONTENT, isPost = true).join()
+            post.reply(AUTHOR, CONTENT).join()
+            post.reply(AUTHOR, CONTENT).join()
+        }
+
+        // Fetch through cache the post
+        val all = cachedForum.root().getAll().join()
+
+        // Reset the forum, acts as a clear of the data
+        resetOnlineForum()
```
Ahayderi, line 136: same comment as above

app/src/androidTest/java/com/github/h3lp3rs/h3lp/forum/ForumCacheTest.kt
```diff
@@ -0,0 +1,166 @@
...
+        val postOnline = cachedForum.newPost(AUTHOR, CONTENT, isPost = true).join()
+
+        // Fetch it again
+        cachedForum.getPost(postOnline.post.key).join()
+
+        // Reset the forum, acts as a clear of the data
+        resetOnlineForum()
+
+        // Fetch through cache the post
+        val all = cachedForum.getAll().join()
```
Ahayderi, line 81: This only seems to work as if the online forum's getAll returns an empty list, you take the value from getAllAux, but if the forum's getAll had returned exceptionally, I think this wouldn't work (c.f. my comment on getAll in the CachedForum)

Ahayderi (approved):
Well, I don't think I've ever taken so much time on a PR review :sweat_smile: and for good reason, this isn't a trivial implementation but I think you did a great job of it! There's a lot of very cool code in here and the cache testing is very interesting.

I've left quite a few comments, some are somewhat problematic (particularly those on the fact that the cached forum doesn't seem to work on all methods if the user is offline) but should be quick fixes.

app/src/main/java/com/github/h3lp3rs/h3lp/forum/implementation/CachedForum.kt
```diff
@@ -0,0 +1,172 @@
...
+ * NOTE: This cache behaves according to the ForumProtocol.md limitations (subset of FireForum
+ * interface)
+ * @param forum The wrapped forum we want to cache
+ */
+class CachedForum(private val forum: Forum) : Forum {
+    override val path = forum.path
+    private val cache = storageOf(FORUM_CACHE)
+
+    // Entry for all the posts at one path
+    data class CacheEntry(val posts: ArrayList<ForumPostData>)
```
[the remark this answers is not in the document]
AlSchlo: Negative, this does not work for serialization

app/src/main/java/com/github/h3lp3rs/h3lp/forum/implementation/CachedForum.kt
```diff
@@ -0,0 +1,172 @@
...
+            }.collect(toList())
+
+            for (categoryPosts in cachedList) {
+                val subForum = CachedForum(forum.root().child(categoryPosts.first))
+                for (post in categoryPosts.second) {
+                    subForum.updateCacheWithPost(emptyList(), post.post)
+                }
+            }
+            // Fetch cache for posts if not found -> If fail we only get an empty list
+            if (cachedList.isEmpty()) getAllAux() else cachedList
```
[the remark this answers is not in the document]
AlSchlo: Great comment! The future always succeeds even if it is offline! This is a design flaw in the SimpleDBForum which was too cumbersome to change. You can see the getAll function of SimpleDBForum and how it simply cannot fail. A bit sad and stupid but that's the reality (I spent a lot of time on trying to change this).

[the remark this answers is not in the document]
AlSchlo: To be more precise, the only thing that can return a failed future in the getAll function of SimpleDBForum is a failed getPost(). However, this does not propagate to the getAll() function properly, since an exception both means: no mapping and/or no connection. So the only way to fix this properly is to distinguish between both cases, way to hard and cumbersome for an already long 7h task.

app/src/main/java/com/github/h3lp3rs/h3lp/forum/implementation/CachedForum.kt
```diff
@@ -0,0 +1,172 @@
...
+        }
+    }
+
+    private fun getAllAux(): List<CategoryPosts> {
```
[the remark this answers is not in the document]
AlSchlo: It's already in a future, so no need to wrap it twice

app/src/main/java/com/github/h3lp3rs/h3lp/forum/implementation/CachedForum.kt
```diff
@@ -0,0 +1,172 @@
...
+            updateCacheWithPost(emptyList(), post.post)
+            post
+        }
+    }
+
+    override fun getPost(relativePath: Path): CompletableFuture<ForumPost> {
+        // Cache the post if successfully found
+        return forum.getPost(relativePath).thenApply {
+            val post = ForumPost(CachedForum(it.forum), it.post, it.replies)
+            updateCacheWithPost((path + relativePath).dropLast(1).drop(path.size), post.post)
```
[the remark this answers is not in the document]
AlSchlo: Ahaha, I swear there was a reason behind this at first :smile: 
Shame on me!

app/src/main/java/com/github/h3lp3rs/h3lp/forum/implementation/CachedForum.kt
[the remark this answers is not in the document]
AlSchlo: Mhhh this may work, but ArrayLists are safe values for serialization. 
Probably that would be smarter though if I can find an implemenation of a map that is easily serializable.

app/src/main/java/com/github/h3lp3rs/h3lp/forum/implementation/CachedForum.kt
```diff
@@ -0,0 +1,172 @@
...
+    @RequiresApi(Build.VERSION_CODES.O)
+    override fun newPost(
+        author: String,
+        content: String,
+        isPost: Boolean
+    ): CompletableFuture<ForumPost> {
+        // Cache the post if successfully returned
+        return forum.newPost(author, content, isPost).thenApply {
+            val post = ForumPost(CachedForum(it.forum), it.post, it.replies)
+            updateCacheWithPost(emptyList(), post.post)
```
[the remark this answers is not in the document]
AlSchlo: No, if the online DB rejects the post and it fails, we don't want it added to the cache.

app/src/main/java/com/github/h3lp3rs/h3lp/forum/implementation/CachedForum.kt
AlSchlo: No, the key of a reply is not repliesKey but key.

app/src/androidTest/java/com/github/h3lp3rs/h3lp/forum/ForumCacheTest.kt
AlSchlo: Composing futures in this case is a big mess, and alters the understanding (and would still require join() nevertheless).
If you insist I can write a comment though :-)

app/src/androidTest/java/com/github/h3lp3rs/h3lp/forum/ForumCacheTest.kt
AlSchlo: Not really, since as explained above, an empty forum only returns failures.
There are no differences in our implementation between both disconnection and absence of data. 
So the current version is fine.

app/src/androidTest/java/com/github/h3lp3rs/h3lp/forum/ForumCacheTest.kt
AlSchlo: Indeed, I switched the order of both lines.
Still confident the test will pass, since I spent a lot of time to make this work.

app/src/androidTest/java/com/github/h3lp3rs/h3lp/forum/ForumCacheTest.kt
AlSchlo: Same as above

app/src/androidTest/java/com/github/h3lp3rs/h3lp/forum/ForumCacheTest.kt
AlSchlo: No, as already explained this is not the case.

app/src/androidTest/java/com/github/h3lp3rs/h3lp/forum/ForumCacheTest.kt
AlSchlo: As discussed, this is not the case

app/src/androidTest/java/com/github/h3lp3rs/h3lp/forum/ForumCacheTest.kt
AlSchlo: I found a critical bug due to this. THANK YOU! :heart:

app/src/main/java/com/github/h3lp3rs/h3lp/forum/implementation/CachedForum.kt
```diff
@@ -0,0 +1,178 @@
...
+        cache.setObject(cachePath, CacheEntry::class.java, entry)
+    }
+
+    override fun listenToAll(action: (ForumPostData) -> Unit) {
+        val newWrappedAction: (ForumPostData) -> Unit = {
+            // This is a very hacky but easy way to know from the data, the path to its forum pointer
+            // We should change the listener design to make it cleaner, but for now it works fine
+            // Impossible to listen for replies, would need to heavy design changes
+            val path =
+                listOf(it.category.name) + if (it.isPost) emptyList() else listOf(it.repliesKey)
```
Lindt19, line 159:
```suggestion
                listOf(it.category.name) + (if (it.isPost) emptyList() else listOf(it.repliesKey))

```

Just for more readability

Lindt19 (approved):
Very impressive cache implementation! Well done for the testing part, I can imagine how tough it was to come up with a fully working forum cache. I'm also glad that it enables having an interesting offline mode of the app 👍🏻
Thank you for the heartwarming PR message! It was a big pleasure working with you ❤️

app/src/main/java/com/github/h3lp3rs/h3lp/forum/implementation/CachedForum.kt
```diff
@@ -0,0 +1,172 @@
...
+            }
+        }
+
+        // First occurrence of key
+        posts.add(postData)
+        setCacheEntry(relativePath, CacheEntry(posts))
+    }
+
+    private fun getCacheEntry(relativePath: Path): CacheEntry {
+        val cachePath = (path + relativePath).joinToString(separator = "/")
```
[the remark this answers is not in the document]
AlSchlo: Even better! Protected! This way it is protected!

app/src/main/java/com/github/h3lp3rs/h3lp/forum/implementation/CachedForum.kt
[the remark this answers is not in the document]
AlSchlo: I started changing it but I remember that the reason I did not do it was in reality the fact that all our forum structures takes lists, so I wanted to stay consistent (and don't bother with set to list (and vice versa) transformations.

app/src/main/java/com/github/h3lp3rs/h3lp/forum/implementation/CachedForum.kt
```diff
@@ -0,0 +1,178 @@
...
+            }
+            // Fetch cache for posts if not found -> If fail we only get an empty list
+            if (cachedList.isEmpty()) getAllAux() else cachedList
+        }
+    }
+
+    private fun getAllAux(): List<CategoryPosts> {
+        if (path.isEmpty()) {
+            // In case we are in the root forum, get the CategoryPosts from all categories
+            val ls = mutableListOf<CategoryPosts>()
```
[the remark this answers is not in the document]
AlSchlo: This looks great but does not work 😆

PR #303, opened 2022-05-25
291 293 link buttons and fix signin bug

Lindt19:
To address #291 and #293 I fixed the sign-in bug where users could go back to the sign-in page from the main as well as the sign-out bug. I also fixed a few other bugs linked to the forum categories and back buttons. On top of that, I linked the remaining buttons in the settings activity with relevant activities, including a new one that I created to enable users to send a report for a bug or a suggestion. Finally, I added more overall tests to increase the coverage.

Now the app is fully functional 🥳


Adrinels (approved):
Nice code !
Happy that this old bug is fixed for our presentation

Your design for bug report is really impressive.  just allowed myself to propose you some color adaptation for our app.

More than that I also change the color of the picture on my own so you don't have to open an image editor on your side ; ).
I push it on the branch so you ave access to it.
But it is your choice to keep it or roll it back to the old design.

Once again well play

And now SDP is over that was your last PR 😢

app/src/androidTest/java/com/github/h3lp3rs/h3lp/SideBarTest.kt
```diff
@@ -80,6 +90,38 @@ class SideBarTest : H3lpAppTest() {
...
+    fun clickingOnRateUsIconSendsToRatingPage() {
+        openDrawerLayout()
+        onView(withId(R.id.nav_view))
+            .perform(NavigationViewActions.navigateTo(R.id.nav_rate_us))
+        intended(allOf(hasComponent(RatingActivity::class.java.name)))
+    }
+
     /**
      * dummy function for coverage, will be deleted later
      */
```
wiam8, line 127: You could delete this now, as you added a correct test ;)

app/src/main/java/com/github/h3lp3rs/h3lp/ReportActivity.kt
```diff
@@ -0,0 +1,62 @@
...
+    fun sendReport(view: View) {
+        val category = reportCategoryDropdown.text.toString()
+        val content = reportEditTxt.text.toString()
+        val report = Report(category, content)
+        // The UID might not be the best solution since a user can't give multiple reports. It's because we need
+        // to have a link to the user that has a bug and also to keep consistency with other
+        // implementations and keep the testing part simple since it's not the most important feature of the app.
+        SignInActivity.userUid?.let {
+            Databases.databaseOf(Databases.REPORTS).setObject(it, Report::class.java, report)
+        }
```
wiam8, line 57: It's true that the UID is needed. However, it could be a simple variable to also send to the database, and you can use a unique id for each new suggestion/bug report.

wiam8 (approved):
Good job! The UI for the reports activity is great!

app/src/main/java/com/github/h3lp3rs/h3lp/ReportActivity.kt
Lindt19: Yes, that is the best choice of implementation. However it would complicate testing more than needed, so I just chose to keep it simple

PR #304, opened 2022-05-31
Change Tutorial and Guide flow

emduc:
Changes the behaviour of the app so that the tutorial is opened only for new users. If a user already agreed to the ToS they won't see the slides again. On the other hand, the guide will be shown on every new install.

Note that there are 2 TODOs that I left out because they require Alexandre adaptation on the map of button->goToActivity so that it can be implemented in a clean way.


app/src/main/java/com/github/h3lp3rs/h3lp/MainPageActivity.kt
```diff
@@ -208,11 +209,19 @@ class MainPageActivity : AppCompatActivity(), OnRequestPermissionsResultCallback
...
      */
     private fun startAppGuide() {
-        if (!storage.getBoolOrDefault(GUIDE_KEY, false)) {
-            storage.setBoolean(GUIDE_KEY, true)
+        val signInStorage = storageOf(SIGN_IN)
+        if (!signInStorage.getBoolOrDefault(GUIDE_KEY, false)) {
+            signInStorage.setBoolean(GUIDE_KEY, true)
+
+            // Disable buttons so that the guide can be seen to completion
+            // TODO change buttonToActivity mapping to disable buttons
```
Ahayderi, line 217: The TODO could maybe be removed? (I'm not sure what our stance on disabling buttons during the guide was so Wiam will know more about this than I do)

app/src/main/java/com/github/h3lp3rs/h3lp/MainPageActivity.kt
```diff
@@ -208,11 +209,19 @@ class MainPageActivity : AppCompatActivity(), OnRequestPermissionsResultCallback
...
+
             // Starts the guide of main page buttons. Once it finishes, it shows the
             // prompt for the search bar by executing the showSearchBarPrompt.
-            showButtonPrompt(mainPageButton) { showSearchBarPrompt() }
+            showButtonPrompt(mainPageButtons) { showSearchBarPrompt() }
+
+            // Re-enable the buttons
+            // TODO
```
Ahayderi, line 224: Same comment as above

Ahayderi (approved):
Short and snappy PR as we like them! Great work as usual, I've left a few comments but they may be due to some misunderstanding on my part

PR #307, opened 2022-06-08
278 remove global context

Ahayderi:
This PR tackles the infamous global context and removes it entirely from our code.

In all fairness, I don't feel as though this is a great solution, it requires many methods to pass a context :sweat_smile: which makes for code that is a bit less clean (please Kotlin, give us implicits!!). On the other hand, having a leaked context was not great either so.. compromises.

As you can expect, the PR basically touches all of our files but it is **very** repetitive so I hope that it won't take too much time, the most interesting files are definitely those from Storages (aka, where these global context stories all first started).


AlSchlo (approved):
Very nice, there is really not a lot to be said unless complaining more about the lack of implicits in Kotlin. Thank you for taking care of this task!

app/src/main/java/com/github/h3lp3rs/h3lp/MedicalCardActivity.kt
```diff
@@ -219,7 +218,7 @@ class MedicalCardActivity : AppCompatActivity() {
     /**
      * Load medical card data
```
Lindt19, line 219:
```suggestion
     * Load the medical card's data
```

app/src/main/java/com/github/h3lp3rs/h3lp/firstaid/GeneralFirstAidActivity.kt
```diff
@@ -0,0 +1,93 @@
...
+            }
+        } ?: run {
+            // If the passed intent was invalid, go back to the main page
+            val intent = Intent(applicationContext, MainPageActivity::class.java)
+            startActivity(intent)
+        }
+    }
+
+    /**
+     * @param backButtonId: Id of the button to go back to the activity with all the first aid tips
```
Lindt19, line 85: I think this documentation is missing the other params

Lindt19 (approved):
So glad that the global context is finally removed! Very nice job 👍

PR #308, opened 2022-06-10
change blocking pull to callback

emduc:
addresses a bug for real android phone where the futures never resolve. Making the pull method blocking would then  crash the app.


wiam8 (approved):

Ahayderi (approved):
Thanks for the fast bug fix!

